PR #7, opened 2019-01-11
Normalise border radii across components

rppld:
# Description

This updates the border radii used in the various components and introduces normalised versions for `sm`, `md` and `lg`.

### DIFF
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -6,6 +6,7 @@ import { Spinner } from '../Spinner'
 import {
   color,
   space,
+  radius,
   fontSize,
   fontWeight,
   lineHeight,
@@ -27,7 +28,7 @@ const StyledButton = styled.button`
   background-color: ${color.earth};
   color: white;
   line-height: ${lineHeight.solid};
-  border-radius: ${props => (props.rounded ? space[32] : space[4])};
+  border-radius: ${props => (props.rounded ? space[32] : radius.md)};
   font-size: ${props => (props.size === 'small' ? fontSize[16] : fontSize[18])};
   width: ${props =>
     props.width === 'full'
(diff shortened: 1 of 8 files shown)


src/components/Input/index.js
```diff
@@ -66,7 +67,7 @@ const FieldWrapper = styled.div`
 const Field = styled.input`
   font-family: inherit;
   background-color: ${color.stardust};
-  border-radius: ${props => (props.rounded ? space[32] : space[4])};
+  border-radius: ${props => (props.rounded ? space[32] : radius.md)};
```
glenngijsberts, line 70: `space[32]` can we also change this with a variable?

src/components/Button/index.js
```diff
@@ -27,7 +28,7 @@ const StyledButton = styled.button`
   background-color: ${color.earth};
   color: white;
   line-height: ${lineHeight.solid};
-  border-radius: ${props => (props.rounded ? space[32] : space[4])};
+  border-radius: ${props => (props.rounded ? space[32] : radius.md)};
```
glenngijsberts, line 31: space[32] can we also change this with a variable?

src/components/Input/index.js
rppld: That's basically just to make it rounded. We could also do `100%` here. Don't think it needs to be a variable, since it's either rounded or not. Which can be controlled in the Input component itself.

src/components/Button/index.js
rppld: Same as for the Input above

src/components/Input/index.js
rppld: Actually we can't use `100%`. That would make a weird shape.

glenngijsberts:
Looking good in storybook 🔥

glenngijsberts (approved):

PR #11, opened 2019-01-15
Allow persistent dialogs

rppld:
# Description

This allows for a `persist` prop on Dialogs to prevent them from closing.


RobVermeer (approved):
👍

PR #20, opened 2019-02-04
Update global CSS reset

rppld:
This adds a more opinionated CSS reset than the one we previously used. All default margins,  paddings and a few other properties are now reset to `0`. After working with Solar for a while, I’ve noticed that adding a global reset like that is the more efficient approach. 99% of the time, we don’t need default styles in the UI’s we’re building.

BREAKING CHANGE: Instead of exporting a `<GlobalStyles />` component, Solar now exports a template literal for global styles which can be extended more easily.


lucasfeliciano (approved):
Clever fix!

PR #22, opened 2019-02-05
Address max-height issues with Dialog component

rppld:
# Description

This fixes the annoying `max-height` behaviour with the `<Dialog />` component by using flexbox instead. Also disables pointer-events on the overflow-gradients.

## How to test

- `$ npm run storybook`
- Verify the dialog with long conent behaves more nicely now

### DIFF
--- a/src/components/Dialog/index.js
+++ b/src/components/Dialog/index.js
@@ -37,10 +37,18 @@ const Content = styled.div`
   width: 100%;
   background-color: white;
   border-radius: ${radius.lg} ${radius.lg} 0 0;
+  overflow: hidden;
+  max-height: 80vh;
+  display: flex;
+  flex-direction: column;
 
   @media ${device.mobileL} {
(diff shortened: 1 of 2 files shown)


RobVermeer (approved):

lucasfeliciano (approved):
Question, this was to solve the problem that we saw on the feedback dialog, right?
Do we experience the same thing on mobile? Because if we do, we should do something about that too.

Not related to this PR just bringing the awareness to that

rppld:
> Question, this was to solve the problem that we saw on the feedback dialog, right?
Do we experience the same thing on mobile? Because if we do, we should do something about that too.

Partly. This adds the overflow-gradient onto the dialog body. Before it was on the footer. So in case a dialog didn’t have a footer, it was missing the bottom overflow-gradient as an indicator that there’s "more". In addition to that, this PR increases the max-height of the Dialog to 80% of the viewport height, so it’ll show a lot more content out of the box.

PR #26, opened 2019-02-07
fix(Button): Set default hover text color for the Button component

RobVermeer:
# Description

Changes the default hover text color of the Button component.
## Changes

- [x] Change default hover text color

## How to test

- Checkout this branch
- `$ npm run docs`


rppld (approved):
👌👌👌

PR #33, opened 2019-02-12
Refactor dialog

rppld:
# Description

This adds support for an optional prop `lockBodyScroll` that if `true` applies `overflow: hidden;` on the `<body>`. Also added a new export `useDialog()` – see the newly added story for how to use it.

### DIFF
--- a/src/components/Dialog/index.js
+++ b/src/components/Dialog/index.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react'
+import React, { Component, useState, useEffect, useRef } from 'react'
 import styled from 'styled-components'
 import { Transition } from 'react-spring'
 import PropTypes from 'prop-types'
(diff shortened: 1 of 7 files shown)


src/components/Dialog/index.js
```diff
@@ -1,4 +1,4 @@
-import React, { Component } from 'react'
+import React, { Component, useState, useEffect, useLayoutEffect } from 'react'
```
lucasfeliciano: 😍

src/components/Dialog/index.js
```diff
@@ -236,76 +250,112 @@ export class Dialog extends Component {
   }
 }
 
-export class DialogWindow extends Component {
-  static propTypes = {
-    on: PropTypes.bool.isRequired,
-    hide: PropTypes.func.isRequired,
-  }
+export function useDialog(props = {}) {
```
lucasfeliciano, line 244: 😍

lucasfeliciano (approved):

RobVermeer:
One issue, when opening the Dialog there is no padding added to the body, thus it can happen that the content moves when opening and closing the dialog.

RobVermeer (approved):
👍

PR #34, opened 2019-02-13
Refactor Box component usage in Banner and Card

rppld:
This replaces the `<Box />` component in two components that had used it.


RobVermeer (approved):

lucasfeliciano:
Let's wait a little bit more to merge this.
I think the memory leak is fixed and my analysis was wrong.
The snapshot that I analyzed was while a request was live, which means that the memory of the renderer was not cleaned yet.
If that is the case, I'm sorry to making you implement this

rppld:
I still think it’s a good idea to remove it anyway. Makes the markup a lot tidier imo. cc @RobVermeer

lucasfeliciano (approved):

PR #39, opened 2019-02-20
Change async arrow fn to avoid bug in Safari

rppld:
Safari has a bug that’ll throw an error if it encounters async arrow function. This PR addresses the issue.

https://github.com/babel/babel/issues/9465
https://github.com/mobxjs/mobx/issues/1169#issuecomment-330060296
https://bugs.webkit.org/show_bug.cgi?id=166879

### DIFF
--- a/src/components/Image/index.js
+++ b/src/components/Image/index.js
@@ -34,6 +34,11 @@ const StyledImage = styled.img`
 `
 
 export class Image extends Component {
+  constructor(props) {
+    super(props)
+    this.handleLoad = this.handleLoad.bind(this)
+  }
+
   observer = null
   initialState = {
     loaded: false,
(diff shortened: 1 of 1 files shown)


src/components/Image/index.js
```diff
@@ -60,10 +60,10 @@ export class Image extends Component {
     return `data:image/svg+xml,%3Csvg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 ${width} ${height}"%3E%3C/svg%3E`
   }
 
-  handleLoad = async () => {
+  async handleLoad() {
```
lucasfeliciano, line 68: I think you need to bind it in the constructor then, otherwise this.setState will be undefined

rppld: Worked fined without doing that, but changed it. Fine for now, maybe gonna refactor this component to use Hooks instead this Friday ✌️

lucasfeliciano (approved):

PR #44, opened 2019-02-26
fix(Alert): Fix spreading of all props to sub components

RobVermeer:
When spreading all the props to sub components styled css would be spreaded to sub components as
well, which lead to css bugs.


rppld (approved):
Thanks 👍

rppld (approved):

PR #66, opened 2019-04-09
Add DateInput component

rppld:
# Description

Adds a `<DatePicker />` component.

## How to test

- Checkout this branch
- `$ npm run storybook`


lucasfeliciano (approved):
Pretty nice!

src/components/DateInput/index.js
```diff
@@ -0,0 +1,156 @@
...
+function createDateString(date) {
+  if (!date) return false
+  let d = date
+  if (typeof date.getMonth !== 'function') {
+    d = new Date(date)
+  }
+  let month = '' + (d.getMonth() + 1)
+  let day = '' + d.getDate()
+  const year = d.getFullYear()
+  month = pad(month)
```
RobVermeer, line 73: You can use the native JS `month.padStart(2, '0')` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart)

src/hooks/useDeviceInfo.js
```diff
@@ -0,0 +1,20 @@
+import { isServer } from '../utils'
+
+export function useDeviceInfo() {
+  if (isServer()) return {}
+  const userAgent = window.navigator.userAgent
+  const isAndroid = () => Boolean(userAgent.match(/Android/i))
+  const isIOS = () => Boolean(userAgent.match(/iPhone|iPad|iPod/i))
+  const isOpera = () => Boolean(userAgent.match(/Opera Mini/i))
+  const isWindows = () => Boolean(userAgent.match(/IEMobile/i))
```
RobVermeer: `isWindowsPhone`? Since it's not just Windows 🤓

RobVermeer:
Why is a select actually an input? 🤔 One issue with that is you can select text inside an input, which looks weird.

Screenshot 2019-04-11 14 07 16

rppld:
> Why is a select actually an input?

Because it behaves the same way as our input components when you focus them, so I used an input. Also because our Input has all the logic to handle and position icons, which I’ve used to add the little arrow on the right.

I added aria attributes based on https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html to make screenreaders interpret this element as an actual dropdown, so it’s also accessible :)

> One issue with that is you can select text inside an input, which looks weird.

Think I can disable that easily with `user-select` or so. Gonna do that!

src/hooks/useDeviceInfo.js
rppld: Good one!

src/components/DateInput/index.js
rppld: Haha I looked at String methods, but that one’s not supported in IE :P

RobVermeer:
1. When hovering the arrow of a select it doesn't show the cursor pointer
2. Pressing spacebar should select the option
3. Nice to have, when typing the letter it would focusses on the first of that letter
4. Select should be a select on mobile, for native UX

rppld:
> When hovering the arrow of a select it doesn't show the cursor pointer ✅
> Pressing spacebar should select the option ✅

> Nice to have, when typing the letter it would focusses on the first of that letter

True, let me think about that one. Might add it in a separate PR.

> Select should be a select on mobile, for native UX

Separate PR :)

src/components/DateInput/index.js
```diff
@@ -0,0 +1,172 @@
...
+  const months = createSelectOptions(props.months)
+  const [initialYear, initialMonth, initialDay] = splitDateString(
+    createDateString(props.initialSelectedDate)
+  )
+  const [year, setYear] = React.useState(initialYear || '')
+  const [month, setMonth] = React.useState(initialMonth - 1 || 0)
+  const [day, setDay] = React.useState(initialDay || '')
+
+  React.useEffect(() => {
+    props.onChange(dateInputRef.current.value)
```
RobVermeer, line 99: Should it always fire the onChange, even when the input field is not valid? 🤔

RobVermeer (approved):
One small comment, rest looks good =)

src/components/DateInput/index.js
rppld: Thought about that too, yeah. Think it should always fire, similar to an `onChange` prop on an `<input />` element. I see validation more as a concern of the form that contains this input.

PR #74, opened 2019-04-18
Disable native date input for Android

rppld:
Support for native date inputs on Android isn’t the greatest yet, so we’re going to show the custom one there.


RobVermeer (approved):
Way easier =)

PR #91, opened 2019-05-14
refactor(Dialog): Update overflow-behaviour

rppld:
This is updating the overflow-behaviour for dialogs on desktop viewports. Dialogs with over-sized contents will now scroll inside the semi-transparent backdrop, instead of in itself. The mobile behaviour remains the same.


RobVermeer (approved):
👌

PR #102, opened 2019-05-17
With Comets

rppld:
# ☄️

This replaces the Icon component with Comets.

### DIFF
--- a/src/components/Icon/index.js
+++ b/src/components/Icon/index.js
@@ -426,24 +426,33 @@ export const Glyph = ({ handle }) => {
   }
 }
 
-export const Icon = ({ size = 32, glyph, ...props }) => (
-  <SvgWrapper size={size} className="icon" {...props}>
-    <InlineSvg
-      fillRule="evenodd"
-      clipRule="evenodd"
-      strokeLinejoin="round"
-      strokeMiterlimit="1.414"
-      xmlns="http://www.w3.org/2000/svg"
-      role="presentation"
-      focusable="false"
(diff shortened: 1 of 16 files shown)


src/components/Icon/index.js
```diff
@@ -426,24 +426,32 @@ export const Glyph = ({ handle }) => {
...
-      viewBox="0 0 32 32"
-      preserveAspectRatio="xMidYMid meet"
-      fit
-    >
-      <Glyph handle={glyph} />
-    </InlineSvg>
-  </SvgWrapper>
-)
+export const Icon = ({ size = 32, glyph, ...props }) => {
+  React.useEffect(() => {
```
RobVermeer, line 430: Can we do this in dev mode only? 🤔

rppld: Good one! Fixed and force-pushed 👍

src/components/Icon/index.js
```diff
@@ -426,24 +426,33 @@ export const Glyph = ({ handle }) => {
...
+    process.env.NODE_ENV === 'development' &&
```
RobVermeer, line 431: Shouldn't we remove the whole effect if not dev, since it's empty for not dev 🤔

rppld: Yeah but you can’t run a hook conditionally. Think it’s fine until we remove the Icon component. Otherwise I might consider removing that warning, lol 😅

RobVermeer (approved):

PR #137, opened 2019-05-31
Allow image-object for responsive sizes

rppld:
* `$ yarn dev`
* http://localhost:9001/?path=/story/cover--full-height

### DIFF
--- a/src/components/Cover/index.js
+++ b/src/components/Cover/index.js
@@ -18,7 +18,7 @@ const Container = styled.div`
   background-color: ${color.space};
   width: 100%;
   height: ${props => (props.height === 'full' ? '100vh' : 'auto')};
-  color: ${props => (props.theme === 'dark' ? 'white' : 'inherit')};
+  color: white;
 `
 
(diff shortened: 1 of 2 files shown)


src/components/Cover/index.js
```diff
@@ -137,12 +173,23 @@ export const Cover = ({
...
+      blurred={blurred}
+      tinted={tinted}
+    />
   </Container>
 )
 
 Cover.propTypes = {
-  imageUrl: PropTypes.string.isRequired,
+  imageUrl: PropTypes.string,
+  images: PropTypes.shape({
```
lucasfeliciano, line 203: What you think about  adding a fallback to the images from the object, otherwise, we will always have to pass all URLs

luciatoma (approved):
Looks good 👍

src/components/Cover/index.js
rppld: Updated the PR to only show images if they’re available.

lucasfeliciano (approved):

PR #213, opened 2019-06-28
fix: add max length on input

luciatoma:
# Description

Added max length property for general Input component.


lucasfeliciano (approved):
Nice !

PR #222, opened 2019-07-05
fix(InputMenuList): Overflow

glenngijsberts:
# Description

Right now the overflow property of the InputMenuList (part of the Input component) has a value of `scroll`. Because of this, windows users (and people who always enabled scroll bars on mac) will see the background of the scrollbar if if the list only contains one item. By using the value `auto` we can prevent this, and if the list get's longer, it will show the scroll bar again.

Not sure if this is the right approach but I created the PR for discussion.

## Changes

- [x] Change overlfow property on InputMenuList

## Screenshots

Before
Schermafbeelding 2019-07-05 om 11 50 11

After
Schermafbeelding 2019-07-05 om 11 49 50

### DIFF
--- a/src/components/Input/index.js
+++ b/src/components/Input/index.js
@@ -171,7 +171,7 @@ export const InputMenuFooter = styled.footer`
 export const InputMenuList = styled.ul`
   max-height: ${space[256]};
   margin: 0;
-  overflow-y: scroll;
+  overflow-y: auto;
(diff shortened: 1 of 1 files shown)


RobVermeer (approved):
Yess! 🙌

rppld (approved):
Awesome! Thx for fixing this.

PR #243, opened 2019-07-17
refactor(Alert): Adjust link-color based on alert-variant

rppld:
Before this change, links in alerts would always be blue. This PR adjust the link colour based on the alert-variant.


luciatoma (approved):
Works great! 👍

PR #264, opened 2019-07-30
fix(MoneyInput): parse initial value with at least 2 decimals

glenngijsberts:
# Description

This will fix an issue we have with the money input, where amounts like 1650 (16,50) will result in 16,5.

## Changes

- [x] Parse amount with fixed decimals of 2 if it's a float.

## How to test

- Checkout this branch
- Checkout the MoneyInput component with the Initial amount

## Screenshots
It will fix this problem:

Schermafbeelding 2019-07-29 om 15 11 17

### DIFF
--- a/src/components/MoneyInput/index.js
+++ b/src/components/MoneyInput/index.js
@@ -69,11 +69,16 @@ export function MoneyInput({
     isCurrencyControlled || props.initialSelectedCurrency
       ? currencies.indexOf(props.currency || props.initialSelectedCurrency)
       : 0
+  const initialValue = props.initialAmount ? props.initialAmount / 100 : 0
+  const [amount, setAmount] = React.useState(
+    props.initialAmount
+      ? Number.isInteger(initialValue)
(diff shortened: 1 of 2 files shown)


src/components/MoneyInput/index.js
```diff
@@ -69,12 +69,17 @@ export function MoneyInput({
     isCurrencyControlled || props.initialSelectedCurrency
       ? currencies.indexOf(props.currency || props.initialSelectedCurrency)
       : 0
+  const initialValue = props.initialAmount ? props.initialAmount / 100 : 0
```
luciatoma, line 72: You can add the property `initialAmount` to the function prop arguments, so that way you don't have to repeat `props.initialAmount` every time, and just use `initialAmount`.

luciatoma (approved):
Nice flow! Also nice that the input has value formatting so you cannot introduce something other than a correct number. 👍
Just left a small comment!

rppld (approved):

PR #273, opened 2019-08-06
With Text and BaseStyles

rppld:
This will add two new components: `<Text />` and `<BaseStyles />`.

`$ yarn dev`


src/components/Text/index.js
```diff
@@ -0,0 +1,55 @@
...
+  hyphens: auto;
+  hyphenate-limit-chars: 6 3 3;
+  hyphenate-limit-lines: 2;
+  hyphenate-limit-last: always;
+  hyphenate-limit-zone: 8%;
+`
+
+export const Text = styled.span`
+  ${baseTextStyles}
+`
```
luciatoma, line 55: Just as an idea, should we define from the start values that can change? For example `fontSize` or `fontWeight`? Maybe we can define some values as default but also give the possibility to pass values via props? It's just a suggestion, don't know the entire context of the scope for this story.

src/components/Text/index.stories.js
```diff
@@ -0,0 +1,18 @@
...
+    Lorem ipsum dolor sit amet consectetur adipisicing elit. Possimus voluptate
+    autem eveniet distinctio dicta magni rem. Atque quaerat obcaecati,
+    cupiditate vel assumenda explicabo aliquid minima inventore tempora tenetur
+    qui modi fuga nisi unde. Quam magni soluta earum rem deleniti, repudiandae
+    sunt, aperiam eaque dolorem odit veniam numquam nulla natus quo incidunt
+    cumque corrupti. Est hic nemo sint reiciendis eum nihil aut esse deleniti
+    illum? Quia architecto nisi enim dignissimos dicta nemo provident explicabo
+    libero ab voluptatem tempora, excepturi nam dolorum dolor cum assumenda
+    nobis quod blanditiis laudantium illum, autem ex laboriosam. Aliquid facilis
+    incidunt maxime veritatis aliquam nulla similique pariatur?
```
luciatoma, line 16: The classical "Lorem ipsum" 😄

lucasfeliciano:
What is the motivation here?

When are we gonna use the `Text` component?

luciatoma (approved):
Looks good! 👍 One comment, not a blocker

PR #285, opened 2019-08-13
refactor(Cover): use picture-element instead of background-image

rppld:
This introduces the `<picture />`-element for our Cover component.

http://localhost:9001/?path=/story/cover--full-height


lucasfeliciano (approved):
Nice one! :)

Roohn (approved):

PR #295, opened 2019-08-21
refactor(dialog): build render-prop API with useDialog

rppld:
# Description

This one has been on my to-do list for a while. Before, logic was duplicated for both the render-prop component `<Dialog />` and the `useDialog()`-hook. Now `<Dialog />` is using `useDialog()` under the hood:

```js
function Dialog({ children, ...props }) {
  const stateAndHelpers = useDialog(props)
  return children(stateAndHelpers)
}
```

😍

## Changes

- [x] Create render-prop API with `useDialog()`
- [x] Adds option to make dialogs controlled

## How to test

- Checkout this branch
- `$ yarn dev`


glenngijsberts (approved):
This looks cool! 🔥

PR #312, opened 2019-09-01
With useTransition hook

rppld:
# Description

This PR adds a custom `useTransition()` hook to the library and replaces `react-spring` in `<Dialog />` and `<MenuButton />`.

## How to test

- Checkout this branch
- `$ yarn dev`
- Visit http://localhost:9001/?path=/story/dialog--basic
- Verify the mount/unmount transition works as expected
- Visit http://localhost:9001/?path=/story/menubutton--default
- Verify the mount/unmount transition works as expected


src/hooks/useTransition.js
```diff
@@ -0,0 +1,58 @@
...
+      }
+      let id = setTimeout(tick, duration)
+      return () => clearTimeout(id)
+    }
+
+    if (on && !prevOn) {
+      setTransitioning(true)
+      setShow(true)
+      requestAnimationFrame(() => {
+        requestAnimationFrame(handleMount)
```
RobVermeer, line 44: Can you explain the need for 2 x `requestAnimationFrame`?

rppld: https://gist.github.com/getify/3002042#gistcomment-2351663

If you’d only call it once, then the initial transition after initial page-load won’t work. It’s bit weird, but you can try it out by removing one call, refresh the page, and open a dialog. Alternatively a `setTimeout()` call with `0` delay also works, but I thought it was uglier than the above.

RobVermeer:
Is it doable to remove all the `react-spring` stuff with this? :)

RobVermeer (approved):
Works as expected, API looks nice! 😄

PR #335, opened 2019-09-09
Refactor Toast component and API

rppld:
Integrates the new useTransition hook and changes the API slightly.

BREAKING CHANGE: Toasts are being called via `notify()` instead of `add()` now. Also exposes a render-callback to remove toasts. See stories for usage examples.


luciatoma (approved):
Looks good! Wondering if we want to add proptypes to these components as well?

rppld:
> Wondering if we want to add proptypes to these components as well?

The `Toast` component accepts only one prop `persist` which there is a check for already.

PR #348, opened 2019-09-20
Allow click and key-handlers for on-controlled dialogs

rppld:
# Description

This allows on-controlled dialogs to be closed with the `Esc`-key and via overlay-click.


glenngijsberts (approved):
Thankyouuuu 🙏

PR #350, opened 2019-09-20
Various visual updates

rppld:
# Description

This adjusts the `lg` radius variable, the Toast background-color, and various Button-states.

## How to test

- Checkout this branch
- `$ yarn dev`

### DIFF
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -11,7 +11,7 @@ import {
   fontSize,
   fontWeight,
   lineHeight,
-  transition,
+  easing,
 } from '../../theme'
 
 const StyledButton = styled.button`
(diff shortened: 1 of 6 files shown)


src/components/Button/index.js
```diff
@@ -51,27 +48,38 @@ const StyledButton = styled.button`
...
+      color: white;
+      text-shadow: ${shadow.text};
+    `};
+
   ${props =>
     props.variant === 'secondary' &&
     css`
-      background-color: ${props.active ? color.earth : color.stardustLight};
-      color: ${props.active ? 'white' : color.space};
+      background-color: ${props.active ? color.earth : '#F0FBFE'};
```
luciatoma, line 72: Just a curiosity here and on line 80 (and other lines where we have colors declared like this), we don't want to add these colors to `theme.js`?

src/components/Button/index.js
```diff
@@ -231,9 +263,52 @@ const StyledButton = styled.button`
...
+      css`
+        background-color: ${color.mars};
+      `};
+
+    ${props =>
+      props.variant === 'secondary' &&
+      props.disabled &&
+      css`
+        color: ${color.earthLighterAlpha};
+        background-color: #f0fbfe;
```
luciatoma, line 296: Maybe add a `cursor: no-drop` for the disabled buttons?

luciatoma (approved):
I love how they look!

src/components/Button/index.js
rppld: We decided against that, because we want to restrict the options developers can choose from. These colours are an exception for these specific button-variants.

src/components/Button/index.js
rppld: Removed that in this PR because the cursor feels huge in Chrome. Buttons by default have the `pointer`-cursor, and disabled buttons just won’t change the cursor on hover. It’s more subtle like that.

PR #359, opened 2019-09-26
Add Panel and Pill components

rppld:
http://localhost:9001/?path=/story/panel--single
http://localhost:9001/?path=/story/panel--group
http://localhost:9001/?path=/story/pill--primary


src/components/Panel/index.stories.js
```diff
@@ -0,0 +1,49 @@
...
+
+export const Group = () => (
+  <Panel>
+    <PanelContent>
+      <PanelBody>
+        <H4>Title</H4>
+        <PanelText>Lorem ipsum dolor sit amet</PanelText>
+        <BaseButton>Edit</BaseButton>
+      </PanelBody>
+      <PanelFooter>
```
glenngijsberts, line 34: Footer? 😛

glenngijsberts:
Don't we need to export this component in `/src/index.js`?

glenngijsberts (approved):
Looks 🔥

src/components/Panel/index.stories.js
rppld: Haha yeahhh, I know. Been thinking about that one for quite a while. Had `PanelAdornment` at first, but it’s not really an "adornment". I think it makes more sense if you look at it this way:

Screenshot 2019-09-26 at 20 44 53

src/components/Pill/index.stories.js
```diff
@@ -0,0 +1,12 @@
...
+
+export default {
+  title: 'Pill',
+}
+
+export const Primary = () => <Pill>Primary</Pill>
+export const Secondary = () => <Pill variant="secondary">Secondary</Pill>
+export const Success = () => <Pill variant="success">Success</Pill>
+export const Danger = () => <Pill variant="danger">Danger</Pill>
+export const Warning = () => <Pill variant="warning">Warning</Pill>
```
rppld, line 13: Add an example with icon and double-check alignment.

PR #404, opened 2019-10-21
Add forward ref to PhoneInput

glenngijsberts:
# Description

This PR will make it possible to focus the PhoneInput on render. If you don't pass a `ref` it will fall back on an internal `ref` so that it will focus the input after selecting a country.

## How to test

- Checkout this branch
- `$ yarn dev`
- Verify that the storybook examples are working as expected


rppld (approved):

luciatoma (approved):
Nice! 👍

PR #420, opened 2019-11-06
HorizontalScroll: Prevent history-navigation

rppld:
🔥 https://dev.to/danburzo/css-micro-tip-prevent-history-navigation-on-horizontally-scrolling-elements-3iil


Roohn (approved):
Nice! Hope that other websites do this as well 😆

PR #438, opened 2019-11-13
fix(Button): change disabled style

luciatoma:
# Description

Improve style for `disabled` button state.
[JIRA task](https://ticketswap.atlassian.net/browse/TTN-1791)

### How to test

- Checkout this branch
- run `yarn`
- Check `disabled` states for `Button` component

### DIFF
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -133,29 +133,8 @@ const StyledButton = styled.button`
   ${props =>
     props.disabled &&
     css`
-      color: rgba(255, 255, 255, 0.6);
-    `};
-
-  ${props =>
-    props.variant === 'secondary' &&
-    props.disabled &&
-    css`
-      color: ${color.earthLighterAlpha};
-    `};
-
-  ${props =>
-    props.variant === 'caution' &&
-    props.disabled &&
-    css`
-      color: ${color.marsLighterAlpha};
-    `};
-
-  ${props =>
-    props.variant === 'inverted' &&
-    props.disabled &&
-    css`
-      background-color: white;
-      color: ${color.earthLighter};
+      opacity: 0.5;
+      pointer-events: none;
     `};
 
   &:focus {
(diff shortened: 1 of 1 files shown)


rppld:
We can basically remove all the specific `disabled` style declarations for the different variants if we move forward with this change. The only one that needs to stay is

```jsx
${props =>
    props.disabled &&
    css`
      opacity: 0.35;
      text-shadow: none;
    `};
```

rppld (changes requested):
We can basically remove all the specific `disabled` style declarations for the different variants if we move forward with this change. The only one that needs to stay is

```jsx
${props =>
    props.disabled &&
    css`
      opacity: 0.35;
      text-shadow: none;
    `};
```

luciatoma:
> We can basically remove all the specific `disabled` style declarations for the different variants if we move forward with this change. The only one that needs to stay is
> 
> ```js-jsx
> ${props =>
>     props.disabled &&
>     css`
>       opacity: 0.35;
>       text-shadow: none;
>     `};
> ```

Updated the PR 👍

src/components/Button/index.js
```diff
@@ -133,29 +133,9 @@ const StyledButton = styled.button`
...
-
-  ${props =>
-    props.variant === 'inverted' &&
-    props.disabled &&
-    css`
-      background-color: white;
-      color: ${color.earthLighter};
+      opacity: 0.35;
+      pointer-events: none;
+      text-shadow: none;
```
rppld: I think the text-shadow can stay for disabled buttons.

src/components/Button/index.js
```diff
@@ -133,29 +133,9 @@ const StyledButton = styled.button`
...
-      color: ${color.marsLighterAlpha};
-    `};
-
-  ${props =>
-    props.variant === 'inverted' &&
-    props.disabled &&
-    css`
-      background-color: white;
-      color: ${color.earthLighter};
+      opacity: 0.35;
```
rppld: I’ll have a look at this with @jelleoverbeek at the office, we might increase this a bit to `0.5`.

rppld (approved):

PR #453, opened 2019-11-22
Add onReset prop for uncontrolled Input components

glenngijsberts:
# Description

Rarely we want to use the Input component as a uncontrolled input. Because this is based on a `ref`, and not value, the `onReset` prop wouldn't work. This PR tries to fix that.

As far as I know a ref isn't reactive so I came up with the solution to always show the clear icon if there is a `onReset` and `ref` prop passed, and then you can conditionally spread the onReset in the parent component. See the added story for more context. 

## Changes

- [x] Changes to onReset
- [x] Added story in storyboard

## How to test

- Checkout this branch
- `$ yarn dev`

### DIFF
--- a/src/components/Input/index.js
+++ b/src/components/Input/index.js
@@ -212,6 +212,9 @@ export const Input = React.forwardRef(
         : typeof validate === 'function'
         ? !validate(props.value)
         : !validate
+    const showResetButton =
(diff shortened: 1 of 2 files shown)


rppld (approved):
👌

PR #468, opened 2019-11-27
Add 'not allowed' cursor to disabled button

frankrrr:
This makes more clear that the button is currently disabled.

hover-disabled

Not sure whether removing `pointer-events: none;` is smart to do.

### DIFF
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -48,7 +48,7 @@ const StyledButton = styled.button`
         ? space[44]
         : space[56]
       : 'auto'};
-  cursor: ${props => (props.disabled ? 'default' : 'pointer')};
+  cursor: ${props => (props.disabled ? 'not-allowed' : 'pointer')};
   height: ${props => (props.size === 'small' ? space[44] : space[56])};
   transition: color 100ms ${easing.easeInOutCubic},
     background-color 100ms ${easing.easeInOutCubic},
@@ -134,7 +134,6 @@ const StyledButton = styled.button`
     props.disabled &&
     css`
       opacity: 0.5;
-      pointer-events: none;
     `};
 
   &:focus {
@@ -186,7 +185,7 @@ const StyledButton = styled.button`
       `};
   }
 
-  &:hover {
+  &:hover:not[disabled] {
     ${props =>
       props.variant === 'primary' &&
       css`


rppld:
We had that in the beginning, but then decided to follow the native implementation for disabled buttons, compare https://codepen.io/rppld/full/ZEEgopm where the cursor doesn’t change either. The `not-allowed` cursor is intended to be used in the context of drag and drop interactions.

Screenshot 2019-11-28 at 13 38 58

From https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

frankrrr:
Okay clear 👍 

What about making these disabled buttons grey? cc @jimzarkadas @jelleoverbeek 

To me it's not super clear that the current buttons are disabled. WDYT?

jimzarkadas:
@frankrrr we recently revamped our disabled buttons and I did some UX research on this topic. It's a bit of controversial topic in general so here is why we designed them this way.

1. We used to have grey disabled button but we changed them to reduced opacity instead
. `Gray doesn’t always represent a disabled condition. Sometimes gray is used to communicate a low priority button in a group (e.g., cancel buttons). Users can easily mistake a grayed out disabled button for a secondary call to action. Or, worst-case scenario, they can mistake it for a poorly designed button with low color contrast.` We also use grey for interactive list items in Solar  (e.g. my tickets for sale page).
More: https://uxmovement.com/buttons/why-you-shouldnt-gray-out-disabled-buttons/
2. `A button that turns from gray to fully colored is an unexpected change that makes users wonder what just happened. For a smooth and seamless experience, it’s best to avoid this. Instead of graying out your disabled buttons, you should decrease the opacity to make it transparent.`
3. Disabled buttons are a bad pattern in general and we want to gradually drop them from our UI. `They don’t give any feedback why they are disabled when you click them.` The right approach to follow is to have the buttons enabled and when the user click on them we show them what's needed to be able to continue. Disabling the action is just a bad common pattern that we are used to in the web world. This is a very nice article about this topic https://axesslab.com/disabled-buttons-suck/

Here are all the disabled states we have. What we could improve instead would be to drop even more the opacity but I personally feel it's pretty ok the way they are at the moment. Did we have any UX complaints about this?

Screenshot 2019-11-29 at 09 43 16
Screenshot 2019-11-29 at 09 43 13
Screenshot 2019-11-29 at 09 43 10

frankrrr:
Thanks for the comprehensive explanation @jimzarkadas, makes a lot of sense to avoid using this pattern in the future! Let's leave it like it is for now.

PR #481, opened 2019-12-04
fix(DateInput): Make it possible to use placeholders for DateInput

glenngijsberts:
# Description

Instead of the init value of the date today we just want to show the placeholders.

## Changes

- [x] Removed prefix of the date when you don't pass init value

## How to test

- Checkout this branch
- `$ yarn dev`
- Checkout the first story

Schermafbeelding 2019-12-04 om 16 03 08

### DIFF
--- a/src/components/DateInput/index.js
+++ b/src/components/DateInput/index.js
@@ -70,10 +70,9 @@ function splitDateString(str) {
   return str.split('-')
 }
 
-function getDate(input) {
+function getDate(input, months) {
   if (!input) {
-    const now = new Date()
-    return [now.getFullYear(), now.getMonth(), now.getDate()]
+    return [null, months[0], null]
   }
   if (typeof input === 'string') {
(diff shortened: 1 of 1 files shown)


src/components/DateInput/index.js
```diff
@@ -73,7 +73,7 @@ function splitDateString(str) {
 function getDate(input) {
   if (!input) {
     const now = new Date()
-    return [now.getFullYear(), now.getMonth(), now.getDate()]
+    return [null, now.getMonth(), null]
```
Roohn: Why not put January as month?

glenngijsberts: oh yeah you're right 😅that was my initial thought as well

glenngijsberts: Fixed!

rppld (approved):
Nice!

PR #482, opened 2019-12-05
DateInput > use null instead of first index of array for months value

glenngijsberts:
# Description

This PR will make sure that the onChange for day and year will work properly without changing the month.

## How to test
- Add a console.log on line `74` for the DateInput component
- The value should become defined as soon as you filled in at least the day and year. Only day should be undefined.
- You can see the difference by checking out master and add the console.log on line `74`


MidasMinnegal (approved):
Nice!

PR #491, opened 2019-12-11
Check for the ref value to show the onReset prop

glenngijsberts:
# Description

Instead of having the logic of showing outside the component, this PR will make it possible to have the logic to show the onReset prop inside the component. See the Storybook story to see how it works without the previous code.

### DIFF
--- a/src/components/Input/index.js
+++ b/src/components/Input/index.js
@@ -212,9 +212,13 @@ export const Input = React.forwardRef(
         : typeof validate === 'function'
         ? !validate(props.value)
         : !validate
-    const showResetButton =
-      (props.value && props.value.length && props.onReset) ||
-      (!props.value && ref && props.onReset)
+    const isControlled = typeof props.value !== 'undefined'
+    const value = isControlled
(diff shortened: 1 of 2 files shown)


src/components/Input/index.js
```diff
@@ -214,7 +214,14 @@ export const Input = React.forwardRef(
...
       (props.value && props.value.length && props.onReset) ||
-      (!props.value && ref && props.onReset)
+      (!props.value &&
+        ref &&
+        props.onReset &&
+        ((ref &&
+          ref.current &&
+          ref.current.value &&
+          ref.current.value.length) ||
+          false))
```
rppld: Bit complex now this part. Maybe add a short comment explaining what’s happening? Controlled with value, uncontrolled with ref and value, etc.

RobVermeer: Force pushed a change, let me know what you think.

RobVermeer (approved):
👌

rppld (approved):

PR #498, opened 2019-12-16
Collapsible with rotate(0deg) instead of rotate()

MidasMinnegal:
# Description
`transform: rotate()` is not a valid `css` attribute so change it into `transform: rotate(0deg)`

## Changes

- [x] Check if we want to rotate and rotate 0deg when we don't want to rotate

## Links
https://github.com/TicketSwap/sirius/pull/1962#discussion_r358321843


rppld (approved):
Great thanks!

glenngijsberts (approved):

PR #514, opened 2019-12-27
Card component updates

rppld:
# Description

Also adjusts line-heights and spacings slightly.

## How to test

- Checkout this branch
- `$ yarn dev`

### DIFF
--- a/src/components/Avatar/index.js
+++ b/src/components/Avatar/index.js
@@ -66,7 +66,13 @@ export const Placeholder = () => (
 export const Avatar = ({ size, src, alt, ...props }) => (
   <Wrapper size={size} {...props}>
     {src ? (
-      <StyledImage src={src} alt={alt} data-testid="image" />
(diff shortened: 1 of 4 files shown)


luciatoma (approved):
Looks really good! 

My only suggestions would be for the `Pill` with the number of available tickets to maybe be centered more with the text? (looks a little bit too high compared with the texts in its left)

Screenshot 2019-12-27 at 14 33 58

Screenshot 2019-12-27 at 14 34 10

rppld:
> My only suggestions would be for the Pill with the number of available tickets to maybe be centered more with the text? (looks a little bit too high compared with the texts in its left)

@luciatoma that’s intentional, see attached screenshot. The vertical alignment of text-content can be controlled with the new `verticalAlign` prop, but in this case it should be aligned at the top.

glenngijsberts (approved):
LGTM!

PR #521, opened 2020-01-08
MoneyInput > Disable select as well when MoneyInput is disabled

glenngijsberts:
# Description

This PR will implement some small changes to the `MoneyInput` component. It will make sure that the currency select is disabled as well when passing the disabled prop. Added a new story in storybook as well.

### DIFF
--- a/src/components/MoneyInput/index.js
+++ b/src/components/MoneyInput/index.js
@@ -62,6 +62,7 @@ export function MoneyInput({
   help,
   validateAmount,
   validateCurrency,
+  disabled,
   ...props
 }) {
   const isCurrencyControlled = typeof props.currency !== 'undefined'
@@ -104,6 +105,7 @@ export function MoneyInput({
                 id={`${inputId}-faux-currency`}
                 label="Currency"
                 hideLabel
+                disabled={disabled}
                 value={
                   isCurrencyControlled ? props.currency.symbol : currency.symbol
                 }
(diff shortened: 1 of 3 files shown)


luciatoma:
Maybe we can add a cursor `no-drop` on the input/select for when `disabled` is `true`? Or add a lighter background color on the input/select. At the moment, besides not being able to click on the inputs, there is no UI to suggest that they are disabled 🤔

src/components/MoneyInput/index.js
```diff
@@ -34,6 +34,10 @@ const SelectWrapper = styled.div`
   }
 `
 
+const StyledSelect = styled(Select)`
+  cursor: default;
+`
```
rppld: I think it’s better to add this logic in the `Select` component instead. Also I assume you only want the default cursor when it’s disabled?

```scss
&:disabled {
  cursor: default;
}
```

glenngijsberts: Ah yes pushed too soon 😅but that is a better solution yes.

rppld:
> Maybe we can add a cursor no-drop on the input/select for when disabled is true?

@luciatoma The no-drop cursor is meant to be used in drag-and-drop kinda UI’s. Think the default cursor is fine if a select is disabled.

glenngijsberts:
> Maybe we can add a cursor `no-drop` on the input/select for when `disabled` is `true`? Or add a lighter background color on the input/select. At the moment, besides not being able to click on the inputs, there is no UI to suggest that they are disabled 🤔

I was thinking the same, but not sure what would be better 🤔Should we ask design?

rppld (approved):
Cool, looks good to me! Think updating the UI of disabled select elements can be done in a separate PR. Shouldn’t be blocking this one, as it’ll also affect input elements I believe.

PR #646, opened 2020-03-04
Add event argument on Toggle component

luciatoma:
# Description

Added the event argument in order  to use the `Toggle` component inside a link component and be able to prevent the parent link action.

### DIFF
--- a/src/components/Toggle/index.js
+++ b/src/components/Toggle/index.js
@@ -22,7 +22,10 @@ export function Toggle(props) {
           ref={checkboxRef}
           type="checkbox"
           defaultChecked={getOn()}
-          onChange={e => props.onToggle(e.target.checked)}
+          onChange={e => {
+            e.preventDefault()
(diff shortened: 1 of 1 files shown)


src/components/Toggle/index.js
```diff
@@ -22,7 +22,7 @@ export function Toggle(props) {
           ref={checkboxRef}
           type="checkbox"
           defaultChecked={getOn()}
-          onChange={e => props.onToggle(e.target.checked)}
+          onChange={e => props.onToggle(e.target.checked, e)}
```
RobVermeer: Why not just do 

```js
onChange={e => {
  e.preventDefault()
  props.onToggle(e.target.checked)
}
```

luciatoma: Yeah I like this idea!

RobVermeer (approved):
👌

PR #659, opened 2020-03-09
fix(moneyinput): when controlled input, let the user parse the amount

RobVermeer:
BREAKING CHANGE: The MoneyInput will now let the parsing of the amount to the parent when it's controlled. This is because otherwise you can't set
amounts in cents. If it's controlled the parent is in full control of what to show.

# Description
When using the MoneyInput controlled, you were not able to change the values after the comma. This PR fixes that issue by making sure the parent is in full control over the value.

## How to test
- Check MoneyInput storybook
- See all work as expected


luciatoma (approved):
Works as expected 👍

Roohn (approved):
Works!

PR #661, opened 2020-03-10
feat: Add topLeftAdornment to Card component

luciatoma:
# Description

Add new prop `topLeftAdornment` to `Card` component


RobVermeer (approved):
👌

PR #667, opened 2020-03-16
fix(portal): the portal showed an error when server side rendered

RobVermeer:
# What it should do
This fixes an issue with server side rendered portals, it should skip it altogether.

## What has been done and why?
- [x] Don't show portal when server side rendered


Roohn (approved):
LGTM!

luciatoma (approved):

PR #800, opened 2020-06-05
With tooltip

rppld:
http://localhost:9001/?path=/story/tooltip--basic


src/components/Tooltip/index.js
```diff
@@ -0,0 +1,81 @@
...
+  height: 0;
+  border-left: 10px solid transparent;
+  border-right: 10px solid transparent;
+  border-bottom: 10px solid ${color.spaceMedium};
+`
+
+export function Tooltip({ children, label, 'aria-label': ariaLabel }) {
+  // get the props from useTooltip
+  const [trigger, tooltip] = useTooltip()
+  // destructure off what we need to position the triangle
```
Roohn: Yes, I can see that 😉

Roohn (commented):
image
Maybe add something that will make the triangle not go to the very outer edge?

Roohn (approved):
One thing, but good for me!

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,81 @@
...
+        // so that collisions don't have a triangle pointing off to nowhere.
+        // Using a Portal may seem a little extreme, but we can keep the
+        // positioning logic simpler here instead of needing to consider
+        // the popup's position relative to the trigger and collisions
+        <Portal>
+          <Triangle
+            style={{
+              left:
+                triggerRect && triggerRect.left - 10 + triggerRect.width / 2,
+              top: triggerRect && triggerRect.bottom + window.scrollY,
```
RobVermeer: Besides a number, what can `triggerRect` be?

If you always want a number here I think you should use parentheses, like: `(triggerRect && triggerRect.bottom) + window.scrollY`

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,81 @@
...
+        // The Triangle. We position it relative to the trigger, not the popup
+        // so that collisions don't have a triangle pointing off to nowhere.
+        // Using a Portal may seem a little extreme, but we can keep the
+        // positioning logic simpler here instead of needing to consider
+        // the popup's position relative to the trigger and collisions
+        <Portal>
+          <Triangle
+            style={{
+              left:
+                triggerRect && triggerRect.left - 10 + triggerRect.width / 2,
```
RobVermeer: Same as other comment and maybe make a const out of this so we can explain why the `10` for example.

src/components/Tooltip/index.js
rppld: > what can triggerRect be?

It’s the return value of `getBoundingClientRect()`.

rppld: Don’t see why the parentheses would be needed?

rppld: Docs are here https://reacttraining.com/reach-ui/tooltip

RobVermeer (approved):

PR #806, opened 2020-06-08
Implement new cover style

Roohn:
# Description
The cover component will not center the content anymore to give some more flexibility for the developer as to how the content will be displayed.
The styling of the cover has been changed slightly.

## Screenshots
image
image

### DIFF
--- a/src/components/Cover/index.js
+++ b/src/components/Cover/index.js
@@ -8,7 +8,6 @@ import {
   fontSize,
   fontWeight,
   transition,
-  lineHeight,
   device,
 } from '../../theme'
 
@@ -17,8 +16,9 @@ const Container = styled.div`
   z-index: 4;
   background-color: ${color.space};
   width: 100%;
-  height: ${props => (props.height === 'full' ? '100vh' : 'auto')};
+  height: ${props => (props.fullHeight ? '100vh' : 'auto')};
   color: white;
+  padding: ${space[16]};
 `
 
 const BackgroundImage = styled.div`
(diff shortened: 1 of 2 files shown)


src/components/Cover/index.stories.js
```diff
@@ -24,24 +25,18 @@ export default {
...
 }
 
-export const basic = () => (
-  <Cover images={data.randomBackgroundImage.imageSizes}>
-    <H1>Awakenings Festival 2019</H1>
-  </Cover>
-)
+const StyledCover = styled(Cover)`
+  text-align: center;
+`
```
RobVermeer: I would not style the storybook, that will make it harder to see the intentional design and usage.

src/components/Cover/index.js
```diff
@@ -208,4 +193,6 @@ Cover.propTypes = {
   }),
   caption: PropTypes.string,
   captionUrl: PropTypes.string,
+  blurred: PropTypes.bool,
+  height: PropTypes.oneOf(['full']),
```
RobVermeer: I think this prop isn't used anymore, right?

If it's still used, then it's super weird that it should be one of one 😄

src/components/Cover/index.js
```diff
@@ -208,4 +193,6 @@ Cover.propTypes = {
   }),
   caption: PropTypes.string,
   captionUrl: PropTypes.string,
+  blurred: PropTypes.bool,
```
RobVermeer, line 224: Best to default to false in this case, so you check on bool.

rppld:
I think if you made the element that has the gradient applied a bit higher, then it will look a bit softer. Right now the color change is quite instant. Know what I mean?

Screenshot 2020-06-08 at 15 10 38

rppld:
I think the caption should be vertically centered and rotated by 90 degrees in the new design. Compare with the screenshot from Slack.

Screenshot 2020-06-08 at 15 12 50

rppld (approved):
Already approving, but I think the caption should be closer to the right edge of the screen. Same distance as it was before towards the bottom.

Screenshot 2020-06-08 at 15 12 50

src/components/Cover/index.js
```diff
@@ -17,137 +16,154 @@ const Container = styled.div`
...
   background-color: ${color.space};
   width: 100%;
-  height: ${props => (props.height === 'full' ? '100vh' : 'auto')};
+  height: ${props => (props.fullHeight ? '100vh' : 'auto')};
   color: white;
+  padding: ${space[16]};
 `
 
 const BackgroundImage = styled.div`
+  filter: ${props => props.blurred && 'blur(15px)'};
```
RobVermeer: When not blurred this would leave an empty value. We should return css when blurred is true, otherwise, nothing.

RobVermeer:
The updated caption results in scrollbars inside the container.

RobVermeer (approved):

Issue #536, opened 2020-01-15
Use @reach/dialog as a base for Dialog


RobVermeer:
See https://reacttraining.com/reach-ui/dialog/

glenngijsberts:
Let's have a discussion about this in our upcoming front-end meeting. Do we want to use this, or do we want to improve the accessibility ourself for example?

PR #864, opened 2020-07-08
OnboardingDialog

glenngijsberts:
# Description

@jelleoverbeek presented a new dialog design to use for forced onboarding. 

## Changes

- [x] Add new component
- [x] Add story

## How to test

- Checkout this branch
- `$ yarn dev`
- Checkout the story 'OnboardingDialog'.

## Screenshots

Example usage:

Screenshot 2020-07-15 at 10 58 06

### DIFF
--- a/src/components/ContentTransition/index.js
+++ b/src/components/ContentTransition/index.js
@@ -90,7 +90,11 @@ export function ContentTransition({ children, onChange, activeView }) {
   )
 }
 
+ContentTransition.defaultProps = {
+  onChange: () => null,
+}
+
 ContentTransition.propTypes = {
   activeView: PropTypes.string.isRequired,
-  onChange: PropTypes.func.isRequired,
+  onChange: PropTypes.func,
 }
--- a/src/components/OnboardingDialog/index.js
+++ b/src/components/OnboardingDialog/index.js
@@ -1,0 +1,151 @@
+import React, { useState } from 'react'
+import styled from '@emotion/styled'
+import PropTypes from 'prop-types'
+import {
+  radius,
+  shadow,
+  space,
+  duration,
+  transition,
+  color,
+  device,
+} from '../../theme'
+import { useTransition } from '../../hooks'
+import { EXITED, ENTERING, ENTERED, EXITING } from '../../hooks/useTransition'
+
+const BodyWrapper = styled.div`
+  background-color: white;
+  border-radius: ${radius.lg};
+  box-shadow: ${shadow.strong};
+  padding: ${space[32]};
+`
+
+const OnboardingDialogBody = ({ children, ...props }) => (
+  <BodyWrapper {...props}>{children}</BodyWrapper>
+)
+
+OnboardingDialogBody.propTypes = {
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.node,
+  ]).isRequired,
+}
+
+const DialogWrapper = styled.div`
+  position: absolute;
+  top: ${space[56]};
+  left: 0;
+  right: 0;
+  z-index: 1;
+
+  @media ${device.tablet} {
+    top: ${space[48]};
+  }
+`
+
+const OnboardingDialog = ({ children, show, ...props }) => {
+  if (!show) return null
+
+  return (
+    <DialogWrapper aria-modal="true" role="dialog" tabindex="-1" {...props}>
+      {children}
+    </DialogWrapper>
+  )
+}
+
+OnboardingDialog.propTypes = {
+  show: PropTypes.bool.isRequired,
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.node,
+  ]).isRequired,
+}
+
+const Container = styled.div`
+  position: relative;
+`
+
+const Backdrop = styled.div`
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-color: ${color.novaLightAlpha};
+  opacity: ${props =>
+    props.state === ENTERING || props.state === ENTERED ? 1 : 0};
+  transition: opacity ${transition};
+`
+
+const OnboardingWrapper = ({ children, active, state, ...props }) => {
+  return (
+    <Container {...props}>
+      {children}
+      {active && <Backdrop state={state} />}
+    </Container>
+  )
+}
+
+OnboardingWrapper.propTypes = {
+  active: PropTypes.bool.isRequired,
+  state: PropTypes.oneOf([EXITED, ENTERING, ENTERED, EXITING]).isRequired,
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.node,
+  ]).isRequired,
+}
+
+const Onboarding = ({ children, defaultOn }) => {
+  const onboardingProps = useOnboarding({
+    defaultOn,
+  })
+
+  return children(onboardingProps)
+}
+
+Onboarding.defaultProps = {
+  defaultOn: true,
+}
+
+Onboarding.propTypes = {
+  defaultOn: PropTypes.bool.isRequired,
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.node,
+  ]).isRequired,
+}
+
+function useOnboarding(config = {}) {
+  const { defaultOn = true, defaultTransitionState = EXITED } = config
+
+  const [show, setShow] = useState(defaultOn)
+  const [active, setActive] = useState(defaultOn)
+  const [state] = useTransition({
+    in: show,
+    timeout: duration,
+    onExited: () => setActive(false),
+    defaultTransitionState,
+  })
+
+  return {
+    show: () => {
+      setActive(true)
+      setShow(true)
+    },
+    hide: () => setShow(false),
+    getOnboardingProps: () => ({
+      show,
+    }),
+    state,
+    active,
+  }
+}
+
+export {
+  OnboardingDialog,
+  OnboardingDialogBody,
+  useOnboarding,
+  OnboardingWrapper,
+}
+
+export default Onboarding
--- a/src/components/OnboardingDialog/index.stories.js
+++ b/src/components/OnboardingDialog/index.stories.js
@@ -1,0 +1,108 @@
+import React, { useState } from 'react'
+import { Button } from '../Button'
+import { H1, H3 } from '../Heading'
+import { Input } from '../Input'
+import styled from '@emotion/styled'
+import { space } from '../../theme'
+import { Text } from '../Text'
+import {
+  OnboardingDialog,
+  OnboardingDialogBody,
+  OnboardingWrapper,
+  useOnboarding,
+} from './index'
+import { ContentTransition } from '../ContentTransition'
+
+const Body = styled.div`
+  padding: ${space[32]};
+`
+
+const Form = styled.form`
+  display: grid;
+  grid-gap: ${space[16]};
+  margin: ${space[32]} 0;
+`
+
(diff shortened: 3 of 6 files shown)


src/components/ContentTransition/index.js
```diff
@@ -90,7 +90,11 @@ export function ContentTransition({ children, onChange, activeView }) {
...
 }
 
+ContentTransition.defaultProps = {
+  onChange: () => null,
+}
+
 ContentTransition.propTypes = {
   activeView: PropTypes.string.isRequired,
-  onChange: PropTypes.func.isRequired,
+  onChange: PropTypes.func,
```
RobVermeer, line 99: This change should be its own commit.

src/components/OnboardingDialog/index.js
```diff
@@ -0,0 +1,139 @@
...
+const OnboardingDialogBody = ({ children }) => (
+  <BodyWrapper>{children}</BodyWrapper>
+)
+
+OnboardingDialogBody.propTypes = {
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.node,
+  ]).isRequired,
+}
```
RobVermeer, line 32: I don't think we need propType validation for children 🤔

src/components/OnboardingDialog/index.js
```diff
@@ -0,0 +1,139 @@
...
+
+const DialogWrapper = styled.div`
+  position: absolute;
+  top: ${props => space[props.positionTop]};
```
RobVermeer: Position top is also dependant on screen-size, so maybe remove the prop, and just do fixed for mobile/desktop 🤔

src/components/OnboardingDialog/index.js
```diff
@@ -0,0 +1,139 @@
...
+  positionTop: 56,
+}
+
+OnboardingDialog.propTypes = {
+  positionTop: PropTypes.number,
+  show: PropTypes.bool.isRequired,
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.node,
+  ]).isRequired,
```
RobVermeer, line 61: Same here, not needed

src/components/OnboardingDialog/index.js
```diff
@@ -0,0 +1,139 @@
...
+  position: relative;
+`
+
+const Backdrop = styled.div`
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-color: rgba(255, 255, 255, 0.8);
```
RobVermeer: Dow we already have something similar, opacity-wise? 🤔 Or maybe this should be within the theme.

src/components/OnboardingDialog/index.js
```diff
@@ -0,0 +1,139 @@
...
+  )
+}
+
+OnboardingWrapper.propTypes = {
+  active: PropTypes.bool.isRequired,
+  state: PropTypes.string.isRequired,
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.node,
+  ]).isRequired,
```
RobVermeer, line 95: Same

src/components/OnboardingDialog/index.js
```diff
@@ -0,0 +1,139 @@
...
+    <Container>
+      {children}
+      {active && <Backdrop state={state} />}
+    </Container>
+  )
+}
+
+OnboardingWrapper.propTypes = {
+  active: PropTypes.bool.isRequired,
+  state: PropTypes.string.isRequired,
```
RobVermeer: This should be a `oneOf(['ENTERED', 'xxx etc'])`

src/components/OnboardingDialog/index.js
```diff
@@ -0,0 +1,139 @@
...
+Onboarding.defaultProps = {
+  defaultOn: true,
+}
+
+Onboarding.propTypes = {
+  defaultOn: PropTypes.bool.isRequired,
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.node,
+  ]).isRequired,
```
RobVermeer, line 115: Same

src/components/OnboardingDialog/index.js
```diff
@@ -0,0 +1,139 @@
...
+}
+
+function useOnboarding(config = {}) {
+  const { defaultOn = true, defaultTransitionState = 'exited' } = config
```
RobVermeer: `exited` should be a const, see also other implementations. Maybe we can import them from `useTransition`

RobVermeer: Btw we then could use those consts in the propType validation as stated above as well.

src/components/OnboardingDialog/index.stories.js
```diff
@@ -0,0 +1,108 @@
...
+const Actions = styled.div`
+  display: grid;
+  margin-top: ${space[16]};
+  grid-template-columns: 200px 200px;
+  grid-gap: ${space[8]};
+`
+
+const StorybookDialog = () => {
+  const { show, hide, getOnboardingProps, state, active } = useOnboarding({
+    defaultTransitionState: 'entered',
```
RobVermeer, line 35: Is the prop `defaultTransitionState` really needed, could you explain what it actually does?

src/components/OnboardingDialog/index.js
glenngijsberts: Why not?

glenngijsberts: Since it _is_ a prop, I think it does no harm to add it. Also, it will remind you if you forget to include it. What do you think, @Roohn?

src/components/OnboardingDialog/index.js
glenngijsberts: Isn't dependant on the child content? I've added it to be able to make adjustments to fit it better with underlaying content.

src/components/OnboardingDialog/index.js
glenngijsberts: I'll check it out.

src/components/OnboardingDialog/index.js
Roohn: Well that is a good one, I never do it because the PropTypes are always the same.
But I see no harm in doing this. Maybe it will be good to do this in every component

Roohn:
@glenngijsberts. Recently we added [@reach/dialog](https://reach.tech/dialog). I think that dialog would be good to use for this. Since you only need to add some styling to it and the accessibility would be better.

Roohn:
Also we have this issue https://github.com/TicketSwap/solar/issues/536 
And this would make the issue even bigger 😅

src/components/OnboardingDialog/index.js
glenngijsberts: Yes let's do that!

src/components/OnboardingDialog/index.stories.js
glenngijsberts: In the current implementation the default state is hard coded with 'exited'. This works fine with our default dialogs, but with this specific dialog it will clearly show that the backdrop will fade in a bit after the page load which looks awful.

src/components/OnboardingDialog/index.js
glenngijsberts: I'll add it to the theme. I'll call it `novaAlpha`.

src/components/OnboardingDialog/index.js
RobVermeer: Then we should make it possible to make a styled DialogWrapper. You can do this by spreading the rest props on the wrapper.

glenngijsberts: Resolved by passing props to DialogWrapper so you can use a styled component instead of you want to alter it.

glenngijsberts:
@Roohn I've looked into it and it seemed that it wasn't really fitting the dialog we created. Rob and I decided to implement the same attributes that Reach is using for the dialog (wrapper). The change is included in my latest commit.

src/theme.js
```diff
@@ -12,6 +12,9 @@ export const color = {
   spaceLighterAlpha: 'rgba(0,19,25,0.2)',
   spaceLightestAlpha: 'rgba(0,19,25,0.1)',
 
+  nova: 'rgb(255, 255, 255)',
+  novaAlpha: 'rgba(255,255,255,0.8)',
```
RobVermeer: These are all the colors that are defined in the design system?

glenngijsberts: I'll ask Jelle.

src/hooks/useTransition.js
```diff
@@ -1,13 +1,14 @@
...
 import { usePrevious } from './usePrevious'
 
-const EXITED = 'exited'
-const ENTERING = 'entering'
-const ENTERED = 'entered'
-const EXITING = 'exiting'
+export const EXITED = 'exited'
+export const ENTERING = 'entering'
+export const ENTERED = 'entered'
+export const EXITING = 'exiting'
```
RobVermeer, line 7: Check the rest of the codebase, we can use these constants there as well (f.e. dialog, content transition)

src/components/OnboardingDialog/index.js
```diff
@@ -0,0 +1,156 @@
...
+  return (
+    <DialogWrapper aria-modal="true" role="dialog" tabindex="-1" {...props}>
+      {children}
+    </DialogWrapper>
+  )
+}
+
+OnboardingDialog.defaultProps = {
+  positionTop: 56,
+}
```
RobVermeer: Can be removed

src/components/OnboardingDialog/index.js
```diff
@@ -0,0 +1,156 @@
...
+
+OnboardingDialog.propTypes = {
+  positionTop: PropTypes.number,
```
RobVermeer: Can be removed

src/components/OnboardingDialog/index.js
```diff
@@ -0,0 +1,156 @@
...
+import { EXITED, ENTERING, ENTERED, EXITING } from '../../hooks/useTransition'
+
+const BodyWrapper = styled.div`
+  background-color: white;
+  border-radius: ${radius.lg};
+  box-shadow: ${shadow.strong};
+  padding: ${space[32]};
+`
+
+const OnboardingDialogBody = ({ children }) => (
```
RobVermeer: I would add `...props` here as well, since this is exported and thus needs styled components support.

src/components/OnboardingDialog/index.js
```diff
@@ -0,0 +1,156 @@
...
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-color: ${color.novaAlpha};
+  opacity: ${props =>
+    props.state === ENTERING || props.state === ENTERED ? 1 : 0};
+  transition: opacity ${transition};
+`
+
+const OnboardingWrapper = ({ children, active, state }) => {
```
RobVermeer: I would add `...props` here as well, since this is exported and thus needs styled components support.

src/components/OnboardingDialog/index.js
```diff
@@ -0,0 +1,156 @@
...
+  duration,
+  transition,
+  color,
+  device,
+} from '../../theme'
+import { useTransition } from '../../hooks'
+import { EXITED, ENTERING, ENTERED, EXITING } from '../../hooks/useTransition'
+
+const BodyWrapper = styled.div`
+  background-color: white;
```
RobVermeer, line 17: `color.nova` (maybe good to do this for everything is a separate commit)

glenngijsberts: I would rather do that in another PR.

src/hooks/useTransition.js
glenngijsberts: I'll do this in a separate PR, so I can test everything and make sure it still works 😅

RobVermeer (approved):

Roohn (approved):

PR #949, opened 2020-09-11
feat(accordion): make it possible to collapse accordion items

glenngijsberts:
# Description

This makes it possible to collapse multiple accordion items at once, which we need for the new ticket alerts dialog. Also added a new story to storybook so check that out to see the difference.


RobVermeer (approved):
👌

PR #1018, opened 2020-12-07
fix(cover): don't use z-index for cover

RobVermeer:
# Description
z-index is not needed for the cover component, so we should not use it. It messes up with other
components, the less a component library messes with z-index, the better.

## Changes
- [x] Remove z-index from cover

## How to test
- Checkout this branch
- `yarn storybook`
- See cover still is the same as before


Roohn (approved):
Noice!

glenngijsberts (approved):

RobVermeer:
This doesn't work as expected, will look at this later...

PR #1024, opened 2020-12-16
feat(MenuButton): Use @reach/menu-button

Roohn:
# Description
The menubutton from `@reach` has way beter accessibility and keyboard functionality.
Next to that I tried to make it so this menubutton can be used in more cases than it did before (more styling options)

## Changes
- [x] Added `@reach/menu-button` and used it in our menuButton component.

## How to test
1. `yarn storybook`
2. Check out the menu button
3. You should be able to navigate the menu with your keyboard (wasn't possible before)

## Screenshots
| Without focus | With focus |
| -- | -- |
| image | image

### DIFF
--- a/src/components/MenuButton/index.tsx
+++ b/src/components/MenuButton/index.tsx
@@ -4,14 +4,17 @@ import {
   MenuItem as ReachMenuItem,
   MenuLink as ReachMenuLink,
   MenuProps,
-  MenuListProps,
+  MenuListProps as ReachMenuListProps,
   useMenuButtonContext,
+  MenuItemProps,
+  MenuLinkProps as ReachMenuLinkProps,
 } from '@reach/menu-button'
 import styled from '@emotion/styled'
 import { color, easing, radius, shadow, space } from '../../theme'
 import React from 'react'
 import { css, Global } from '@emotion/react'
 import { useTransition } from '../../hooks'
+import { TransitionState } from '../../hooks/useTransition'
 
 const DURATION = 200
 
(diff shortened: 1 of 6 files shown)


glenngijsberts:
What is the purpose of having the blue focus border? Does this really improve the accessibility? Imo it's a downgrade on aesthetic level. Same story for the lack of mini animations.

src/components/MenuButton/index.tsx
```diff
@@ -1,147 +1,51 @@
...
-  transition: opacity ${duration}ms ease-out, transform ${duration}ms ease-out;
+// @ts-ignore
+export const MenuItem = styled(ReachMenuItem)`
+  padding: ${space[8]} ${space[16]};
+  cursor: pointer;
 `
 
-const Menu = styled.div`
-  display: inline-block;
+// @ts-ignore
```
glenngijsberts: Why do we need the @ ts-ignore?

Roohn:
> What is the purpose of having the blue focus border? Does this really improve the accessibility? Imo it's a downgrade on aesthetic level. Same story for the lack of mini animations.

I agree on the animations, I will try to have those in as well.
As far as the focus border is concerned this is a crucial part for people who rely on it as stated in [this](https://hiddedevries.nl/en/blog/2019-06-06-indicating-focus-to-improve-accessibility) article

glenngijsberts:
Screenshot 2020-12-17 at 09 55 08

@Roohn It looks different on my side 🤔 That's where my question was coming from 😅 if it looks like your screenshot it would be fine.

src/components/MenuButton/index.tsx
Roohn: Very strange type collision between reach and emotion 🤔 
I haven't figured it out yet, so I will make an issue for it

glenngijsberts (dismissed):
Nice 👍 Only thing that's missing is the fade out if you dismiss the menu button, that still looks a little odd.

RobVermeer (dismissed):
Looks good, works well, I only think this is a breaking change =)

Roohn:
I added this as a breaking change @RobVermeer 👍

PR #1057, opened 2021-01-26
Button component prop and styling cleanup

Roohn:
# Description
The button component had some unused and unnecessary props. This component also wasn't in line with our designs anymore because there are three sizes (small, medium, large) instead of two (default, small). 

## Changes
- [x] Removed props: `square` & `rounded`
- [x] Adjusted props: `size` & `width`
- [x] Refactored code

## How to test
There are some components in storybook that use the button so that is a good place to start.

## Screenshots
Sizes:
image
Different states:
image
With adornments:
image

### DIFF
--- a/src/components/Button/index.stories.tsx
+++ b/src/components/Button/index.stories.tsx
@@ -1,210 +1,104 @@
-import React from 'react'
-import { Button, ButtonVariant } from '.'
-import {
-  Checkmark,
-  Facebook,
-  MagnifyingGlass,
-  Plus,
-  Apple,
-} from '@ticketswap/comets'
+import React, { Fragment } from 'react'
+import { Button, ButtonSize, ButtonVariant } from '.'
+import { Alert } from '@ticketswap/comets'
+import styled from '@emotion/styled'
+import { space } from '../../theme'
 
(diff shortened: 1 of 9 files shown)


glenngijsberts:
Really nice work @Roohn! :) a question tho, I see that you removed the active state. Did you do that on purpose and if so, how would you achieve the same result without it? Example below:

Screenshot 2021-01-29 at 11 05 15

glenngijsberts:
Feedback from Jelle: Can you adjust the secondary button to have "SkyLightAlpha" as background color and "SkyAlpha" as hover background color? You can ask Jelle for the values of the Alpha versions.

Roohn:
> Really nice work @Roohn! :) a question tho, I see that you removed the active state. Did you do that on purpose and if so, how would you achieve the same result without it? Example below:
> 
> Screenshot 2021-01-29 at 11 05 15

I removed it because I couldn't find any use of it in the code. And to be honest the name `active` wasn't telling me much, but now I see what it is used for, so I will add it 👍🏼

glenngijsberts (approved):
Really nice work Ronald :)

PR #1129, opened 2021-06-02
Add extra tests to the components

### LINKED ISSUE #1041: Write a test for each component
## Task description
A lot of our components are not having tests. We can definitely improve this by writing at least one test for each component.

## Expected outcome
100% component testing

glenngijsberts:
# Description

I always had the idea of having more tests for our components. I've tried to create tests for most of our components. For some components I've only added the "renders without crashing" test and for some I didn't add a test at all, because I didn't saw any benefits for it (such as the Text and other styled components only).

## How to test

* See the tests passing, checkout the code

## Links

fixes https://github.com/TicketSwap/solar/issues/1041


Roohn (approved):
Awesome work!

PR #1151, opened 2021-07-16
Detect if input menu is partially outside of viewport 

glenngijsberts:
# Description

This PR will make sure that the viewport is taking into account when trying to use the select box.

## How to test

- Checkout the Select story in storybook
- Checkout the toast withInput story in storybook

## Screenshots

Screenshot 2021-07-16 at 18 09 26
Screenshot 2021-07-16 at 18 09 11

### DIFF
--- a/src/components/Input/index.tsx
+++ b/src/components/Input/index.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode, Ref } from 'react'
+import React, { ReactNode, Ref, useRef, useState, useEffect } from 'react'
 import styled from '@emotion/styled'
 import { css } from '@emotion/react'
 import { Spinner } from '../Spinner'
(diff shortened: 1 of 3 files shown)


src/components/Input/index.tsx
```diff
@@ -254,6 +237,70 @@ const InputMenuItemContainer = styled.li<InputMenuItemProps>`
...
+  useEffect(() => {
+    setHeight(ref?.current?.clientHeight || 0)
+  }, [setHeight])
+
+  useEffect(() => {
+    const bottom = ref?.current?.getBoundingClientRect()?.bottom || 0
+    const viewportHeight = window?.innerHeight || 0
+
+    if (bottom > viewportHeight) {
+      setIsOutsideViewport(true)
```
Roohn, line 293: Shouldn't this be updated when scrolling? When a select is in a form at the bottom of a page it will go up. But the form goes further after scrolling, it should go down again.

glenngijsberts: I don't see this as a problem at the moment. Bit of an edge case if users are trying to select something & then trying to scroll to the bottom. Imo it would also look weird if it changes if you focussing the input & you're scrolling.

RobVermeer (approved):
🥇

PR #1187, opened 2021-09-10
Introduce components code samples in Storybook

diogoredin:
# Description

This PR adds a docs add-on to Storybook so we can see each component's code sample. Now developers don't need to clone or open Solar to check how components can be used. In the future, we can improve this further by also adding a table with the component properties and their types.

This PR also replaces the dark mode toggle with the background selector from Storybook, so the selection can be kept between stories and we can remove the minimum height of each story container.

## Changes

- [x] This has been done
- [ ] I still need to to this

In order to do this change, there was a necessary fix so emotion was correctly loaded in Storybook (there was one missing mappings in the Webpack config). Besides this, some components need to now be exported directly (without intermediate variables) so Storybook can correctly find their code.

## How to test

- Checkout this branch
- `$ yarn storybook`

## Screenshots

Screenshot 2021-09-10 at 15 02 56

## To be notified

@bob @glenngijsberts @RobVermeer @Roohn

## Links

The current version (without the source code): https://ticketswap.github.io/solar/


glenngijsberts:
I think it would be nice to use `chore` or `docs` instead of `fix`, since that commit is fixing documentation, rather than components. If you use `fix`, it will publish a new version of the Solar package and that's not something we want if we only change storybook :)

src/components/MenuButton/index.tsx
```diff
@@ -97,6 +97,7 @@ export const MenuItem = styled<React.FC<MenuItemProps>>(ReachMenuItem)`
 
 export interface MenuLinkProps extends ReachMenuLinkProps {
   href: string | undefined
+  as: string | undefined
```
bobbonius, line 100: Isn't it better to make it `nullable` instead of using `undefined`? Same goes for `href`

RobVermeer (approved):

PR #1190, opened 2021-09-20
Introduce theme colors

bobbonius:
# Description

While adding the theme color stories we removed the `focusAlpha` variant. This variant is used for the `focus` state of, for example, a `<Button />` but is too light and therefore gives contrast errors when checking a11y requirements. We now use `lighterAlpha` for `focus` and decided to remove `focusAlpha` as it's unused now and also not consistently added to all colors.

## How to test

- Checkout PR
- `yarn && yarn storybook`
- Go to the `<Button />` component
- Test if the focus color is working and displayed as intended
- Go to a color in `Design Tokens/Colors
- See if the colors are displayed

## Screenshots
128f828b3b772c2d84a9d79759e1e0dd


glenngijsberts (approved):
Nice work Bob 🤙

bobbonius:
Added typing to the `color` object from `theme.ts` as it gave the error below.

ce1f630ccad1d8e17e0787742f25f598

RobVermeer (approved):
😍 Awesome!

PR #1195, opened 2021-09-22
Grouping components into catagories

bobbonius:
# Description

Together with design we added categories so we can group our components.

## How to test

- Checkout this branch
- `yarn storybook`
- Check if the categories are in the menu on the left

## Screenshots
080ec3f26583fa6eff78fa34c3d2ba64

### DIFF
--- a/src/components/Accordion/index.stories.tsx
+++ b/src/components/Accordion/index.stories.tsx
@@ -59,5 +59,5 @@ export const Collapsible = () => (
 )
 
 export default {
-  title: 'Accordion',
+  title: 'Components/Surfaces/Accordion',
 }
--- a/src/components/Alert/index.stories.tsx
+++ b/src/components/Alert/index.stories.tsx
@@ -109,5 +109,5 @@ export const Warning = () => (
 )
 
 export default {
-  title: 'Alert',
+  title: 'Components/Feedback/Alert',
 }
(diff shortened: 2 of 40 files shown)


bobbonius:
We also had a discussion about renaming `BaseButton` to `Link` as we all thought they look more like a link.

Roohn:
> We also had a discussion about renaming `BaseButton` to `Link` as we all thought they look more like a link.

It looks like a link, but is still a button. So I wouldn't call it link to avoid a lot of confusion 😬

glenngijsberts:
I would display the `OnboardingDialog` at the same category as the `Dialog` component. Besides that, the `OnboardingDialog` has a duplicate story.

glenngijsberts:
> > We also had a discussion about renaming `BaseButton` to `Link` as we all thought they look more like a link.
> 
> It looks like a link, but is still a button. So I wouldn't call it link to avoid a lot of confusion 😬

Also it would be conflicting with the `Link` from Next 😬

Roohn:
The following categories are a bit meaningless to me:
- Surfaces
- Data display
- Layout
- Content
Every single component has a surface, displays data, has a layout and has content… 😅 
Could you maybe elaborate on the reason why you choose these?

RobVermeer (approved):
I think we can improve on the categories, but that will come over time, have no suggestions atm! Really nice though! 😄

glenngijsberts (approved):
Nice work 🤙

PR #1196, opened 2021-09-23
Migrate Comets to Solar

glenngijsberts:
# Description

We experienced that maintaining two different repo's (that are essentially the same in terms of Stack) is hard & making things harder for us. The ideal situation would be to include Comets into Solar, but first we need to make sure that we update Solar to another build/bundler tool so that we are able to tree shake better.


glenngijsberts:
Outdated, so I'll close this for now. Let's do this in a new branch/PR.

PR #1198, opened 2021-09-24
docs(toast): refactor toast stories

### LINKED ISSUE #1185: Remove Toast consumer
## Improvement description
We want to remove the Toast consumer because the context is already exposed in the useToast hook. Why are we offering the consumer if we prefer the hook?

https://github.com/TicketSwap/solar/blob/main/src/components/Toast/index.tsx

bobbonius:
# Description

Refactored the toast stories and let them all use the useToast hook (as we think that should be the way to go) instead of getting the notify function from the consumer

## How to test

- Checkout this branch
- `yarn storybook`
- Check out the toast and see if they work

Fixes #1185

### DIFF
--- a/src/components/Toast/index.stories.tsx
+++ b/src/components/Toast/index.stories.tsx
@@ -1,21 +1,29 @@
 import React from 'react'
-import { Toast, ToastConsumer, ToastProvider, useToast } from './'
+import { Toast, ToastProvider, useToast } from './'
 import { Button, ButtonSize, ButtonVariant } from '../Button'
 import { Select } from '../Select'
+import styled from '@emotion/styled'
+import { Text } from '../Text'
+
+const ToastContent = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+`
 
 const Wrapper = (story: () => React.ReactNode) => (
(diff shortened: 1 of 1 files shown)


diogoredin:
Awesome 😁 Just one question, did you consider removing this option of using the notify altogether from the code and force clients to use the hook? At first glance, Sirius seems to have 15 occurrences of this usage.

RobVermeer:
> Awesome 😁 Just one question, did you consider removing this option of using the notify altogether from the code and force clients to use the hook? At first glance, Sirius seems to have 15 occurrences of this usage.

I'm not too sure what would be the best practise here. Users may want to use the consumer instead of the hook, which sounds perfectly fine. But enforcing a best practise sounds good as well 😄

RobVermeer (approved):
👌

bobbonius:
> Awesome 😁 Just one question, did you consider removing this option of using the notify altogether from the code and force clients to use the hook? At first glance, Sirius seems to have 15 occurrences of this usage.

Not really no, but that might be the best thing to do going forward. For now we just wanted to remove the `Consumer` examples from the stories so people only learn about the `useToast` method.

PR #1199, opened 2021-09-24
feat(moneyinput): support for maximumAmount

SavageTiger:
# Description

In order to make it possible to maximize the amount that can be entered in a money field a new
property was added to configure a maximum amount.

## How to test

- Open the storybook
- Goto `With maximum amount` under the `MoneyInput`
- Try to enter a amount bigger then `30`
- Verify other stories related to `MoneyInput` still work as expected
- Verify it works as expected
- Order pizza to celebrate 🍕 

## Screenshots

image

### DIFF
--- a/src/components/MoneyInput/index.tsx
+++ b/src/components/MoneyInput/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 import styled from '@emotion/styled'
 import { space } from '../../theme'
 import { Label, LabelText, Input, Help } from '../Input'
@@ -28,6 +28,7 @@ export interface MoneyInputProps {
   initialSelectedCurrency?: MoneyInputCurrencyProp
   amount?: number | string
   initialAmount?: number
+  maximumAmount?: number
   validateAmount?: boolean
   validateCurrency?: boolean
   disabled?: boolean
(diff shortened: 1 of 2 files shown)


src/components/MoneyInput/index.tsx
```diff
@@ -120,6 +128,29 @@ export const MoneyInput: React.FC<MoneyInputProps> = ({
...
+      parseAmount(amount) > maximumAmount
+    ) {
+      setAmount(maximumAmount / 100)
+    } else {
+      onChange({
+        currency,
+        amount: isAmountControlled ? amount : parseAmount(amount),
+      })
+    }
+  }, [amount])
```
bobbonius, line 152: There are a lot of dependencies missing on the dependency array. Also, the `onChange` being called a lot in the `useEffect` doesn't seem right. Might be nice to use a `useCallback` here like this:

```
  const handleChange = useCallback(
    (event: React.ChangeEvent<HTMLInputElement>) => {
      const amount = event.target.value
      if (
        !isAmountControlled &&
        hasMaximumAmount &&
        parseAmount(amount) > maximumAmount
      ) {
        setAmount(maximumAmount / 100)
      } else {
        setAmount(amount)
      }
    },
    [hasMaximumAmount, isAmountControlled, maximumAmount]
  )
  ```
  
  Then you add this `handleChange` to the onChange prop and it will already generate one less render.

bobbonius (commented):
In general, I don't think it's a good idea to control the user input like you're doing with this feature. There are no warnings whatsoever. If a user inputs like 45 euros and the maximum is set to 30 and doesnt notice this change (and we have a lot of those users at TicketSwap), you will get all sorts of customer support related issues.

src/components/MoneyInput/index.tsx
glenngijsberts: I'll check this out :)

glenngijsberts:
> In general, I don't think it's a good idea to control the user input like you're doing with this feature. There are no warnings whatsoever. If a user inputs like 45 euros and the maximum is set to 30 and doesnt notice this change (and we have a lot of those users at TicketSwap), you will get all sorts of customer support related issues.

@bobbonius context: currently it's possible to fill in a looooooot of numbers in this input. We want to maximize this. Of course we're not gonna set this to 30 euros, but we'll make sure to improve this experience a bit :)

SavageTiger:
A finding we also had today is if you add alphanumeric numbers the check seems to break, so that needs fixing before we can merge this :)

bobbonius:
Closing this PR, might be nice for future endeavors

PR #1210, opened 2021-10-04
Improve Select onKeyDown events

bobbonius:
# Description

A native `<select>` element has the ability to take in input from the user using a keyboard, for example for a11y reasons. The Select component was not able to handle this, only escape / space and key down / key up events. This PR adds support for search inputs in a select dropdown. The query you type in as a user resets after 1 seconds.

I noticed the `<Toast />` component wasn't in a category (must have missed it the other PR) so I also added it to the right category.

## How to test

- Checkout this branch
- `$ yarn storybook`
- Go to `<Select />` component
- See if you can search on the countries in the list by typing in a few letters of the country you're looking for

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "@storybook/storybook-deployer": "^2.8.10",
     "@testing-library/jest-dom": "^5.14.1",
     "@testing-library/react": "^11.2.2",
+    "@testing-library/user-event": "^13.2.1",
     "@ticketswap/comets": "^2.14.0",
     "@types/react": "^17.0.14",
     "@types/react-dom": "^17.0.0",
--- a/package.json
+++ b/package.json
@@ -76,7 +76,6 @@
     "@storybook/storybook-deployer": "^2.8.10",
     "@testing-library/jest-dom": "^5.14.1",
     "@testing-library/react": "^11.2.2",
-    "@testing-library/user-event": "^13.2.1",
     "@ticketswap/comets": "^2.14.0",
     "@types/react": "^17.0.14",
     "@types/react-dom": "^17.0.0",
--- a/src/components/Select/index.tsx
+++ b/src/components/Select/index.tsx
@@ -1,4 +1,13 @@
-import React, { ReactNode } from 'react'
+import React, {
+  ReactNode,
+  useRef,
+  useCallback,
+  useState,
+  useEffect,
+  SyntheticEvent,
+  KeyboardEvent,
+  FC,
+} from 'react'
 import styled from '@emotion/styled'
 import computeScrollIntoView from 'compute-scroll-into-view'
 import { color, space, lineHeight, radius, fontSize } from '../../theme'
@@ -16,6 +25,7 @@ import {
 import { VisuallyHidden } from '../VisuallyHidden'
 import { ArrowDown } from '@ticketswap/comets'
 import { useOnClickOutside, useKeyPress, useDeviceInfo } from '../../hooks'
+import { debounce } from '../../utils/debounce'
 
 const Container = styled.div`
   position: relative;
@@ -104,7 +114,7 @@ const StyledSelect = styled.select<StyledSelectProps>`
   }
 `
 
-export const Select: React.FC<SelectProps> = ({
+export const Select: FC<SelectProps> = ({
   items,
   onChange = () => {},
   id,
@@ -115,27 +125,30 @@ export const Select: React.FC<SelectProps> = ({
   ...props
 }) => {
   const { isMobile } = useDeviceInfo()
-  const [showCustomSelect, setShowCustomSelect] = React.useState(false)
-  const [isOpen, setIsOpen] = React.useState(false)
-  const [selectedItem, setSelectedItem] = React.useState(
+  const [showCustomSelect, setShowCustomSelect] = useState(false)
+  const [isOpen, setIsOpen] = useState(false)
+  const [query, setQuery] = useState('')
+  const [selectedItem, setSelectedItem] = useState(
     initialSelectedItem || items[0]
   )
-  const [highlightedIndex, setHighlightedIndex] = React.useState(
+  const [highlightedIndex, setHighlightedIndex] = useState(
     items.indexOf(initialSelectedItem || items[0])
   )
-  const containerRef = React.useRef<HTMLDivElement>(null)
-  const inputRef = React.useRef<HTMLInputElement>(null)
-  const menuRef = React.useRef<HTMLUListElement>(null)
+  const containerRef = useRef<HTMLDivElement>(null)
+  const inputRef = useRef<HTMLInputElement>(null)
+  const menuRef = useRef<HTMLUListElement>(null)
   let itemRefs: HTMLLIElement[] = []
   const labelId = `${id}-label`
   const menuId = `${id}-menu`
   const esc = useKeyPress('Escape')
   const arrowUp = useKeyPress('ArrowUp')
   const arrowDown = useKeyPress('ArrowDown')
+  const typeAheadTimeout = 1000
+  const resetQuery = debounce(() => setQuery(''), typeAheadTimeout)
 
   useOnClickOutside(containerRef, () => setIsOpen(false))
 
-  const handleClose = React.useCallback(() => {
+  const handleClose = useCallback(() => {
     if (!isOpen) return false
     setIsOpen(false)
 
@@ -149,7 +162,7 @@ export const Select: React.FC<SelectProps> = ({
     return (itemRefs = [...itemRefs, el])
   }
 
-  React.useEffect(() => {
+  useEffect(() => {
     if (!isMobile()) {
       // Component will render a native select element by default.
       // Upon mounting, we check whether we’re on a mobile device. If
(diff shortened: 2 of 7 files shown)


src/components/Select/index.tsx
```diff
@@ -292,27 +348,7 @@ export const Select: React.FC<SelectProps> = ({
...
-                )
-              }
-              // Select with Enter or Space keys
-              if (e.key === 'Enter' || e.keyCode === 32) {
-                onChange(items[highlightedIndex])
-                setSelectedItem(items[highlightedIndex])
-                handleClose()
-              }
-            }}
+            onKeyDown={handleKeyDown}
```
glenngijsberts, line 354: Nice

src/utils/debounce/index.ts
```diff
@@ -0,0 +1,8 @@
+export const debounce = (callback: Function, delay: number) => {
+  let timeout: number
+
+  return () => {
+    clearTimeout(timeout)
+    timeout = setTimeout(callback, delay)
+  }
+}
```
glenngijsberts: Hmm. I know that we're already having something inside Sirius for debouncing (`useDebounce`). Might be nice to move this to Solar, so we can have one dedicated util/hook for debouncing, and not 1 in Solar and 1 in Sirius 😬

glenngijsberts:
Not sure if I like it how it works after you searched for something. After I typed Dutch I can't type something else, unless I press backspace a few times. Might be nice to search for something else every time after the debounce.

src/components/Select/index.tsx
```diff
@@ -115,27 +125,32 @@ export const Select: React.FC<SelectProps> = ({
...
+  const inputRef = useRef<HTMLInputElement>(null)
+  const menuRef = useRef<HTMLUListElement>(null)
   let itemRefs: HTMLLIElement[] = []
   const labelId = `${id}-label`
   const menuId = `${id}-menu`
   const esc = useKeyPress('Escape')
   const arrowUp = useKeyPress('ArrowUp')
   const arrowDown = useKeyPress('ArrowDown')
+  const resetQuery = useCallback(
+    debounce(() => setQuery(''), 1500),
```
Roohn: I think this is a bit too long. You are not waiting for 1.5 seconds until you type your next letter. Native selects in browsers have a bit shorter delay I think

RobVermeer: 1000 😬

bobbonius: I'll switch this to 1 second as defined in the HTML source code: [typeAheadTimeout](https://chromium.googlesource.com/chromium/blink.git/+/refs/heads/main/Source/core/html/forms/TypeAhead.cpp#45)

src/utils/debounce/index.ts
bobbonius: This debounce works different then the one in Sirius though. Not sure if I want to use that hook in this situation.

bobbonius:
> Not sure if I like it how it works after you searched for something. After I typed Dutch I can't type something else, unless I press backspace a few times. Might be nice to search for something else every time after the debounce.

@glenngijsberts The component doesn't listen for a backspace event. A native `select` element doesn't listen to a backspace event either. It resets the entire query after 1 seconds and that's what I tried to mock here with the debounce. So pressing backspace a few times didn't actually do something but it did pass some time which eventually resets it.

glenngijsberts:
@bobbonius ah right, makes sense. Anyway, still think 1 second is maybe a bit too long although it's the default.

glenngijsberts (approved):

src/components/Select/index.tsx
```diff
@@ -231,6 +247,47 @@ export const Select: React.FC<SelectProps> = ({
...
+  const handleKeyDown = (e: KeyboardEvent) => {
+    if (!isOpen) return false
+
+    const inputIsALetter = /^[a-z\d]$/i.test(e.key)
+    const arrowUp = e.key === 'ArrowUp'
+    const arrowDown = e.key === 'ArrowDown'
+    const enterOrSpace = e.key === 'Enter' || e.code === 'Space'
+
+    if (inputIsALetter) {
+      resetQuery()
```
diogoredin, line 261: **thought/suggestion:** With this approach we have no guarantee that when the component unmounts we have no debounced `resetQuery` to run. If we have one we could get React trying to update the state of an unmounted component.

We could try doing the following:
1. Expose a cancel handler in the debounce function, similar to what `lodash` does (which clears the timeout);
2. Introduce a `useEffect` that:
- listens to changes on the query;
- debouces it's reset;
- when this component is unmounted cancels the debounce (cancel in the return).

src/components/Select/index.tsx
```diff
@@ -231,6 +247,47 @@ export const Select: React.FC<SelectProps> = ({
...
+
+      const input = query + e.key
+      setQuery(input)
```
diogoredin: **thought/suggestion:** We have no guarantee the query is exactly the last one because React might batch these updates. https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

`setQuery(prevQuery => prevQuery + input)` would be more correct I think, and then we would use `query` for the operations below instead of input. This should work since the query itself is not debounced but I didn't test it, so not 100% sure.

src/components/Select/index.tsx
```diff
@@ -231,6 +247,47 @@ export const Select: React.FC<SelectProps> = ({
     </SelectMenu>
   )
 
+  const handleKeyDown = (e: KeyboardEvent) => {
+    if (!isOpen) return false
+
+    const inputIsALetter = /^[a-z\d]$/i.test(e.key)
+    const arrowUp = e.key === 'ArrowUp'
+    const arrowDown = e.key === 'ArrowDown'
+    const enterOrSpace = e.key === 'Enter' || e.code === 'Space'
```
diogoredin: **nit-pick:** I think this would be more easily understandable if all variables followed the same style (`inputIsArrowUp`, `inputIsArrowDown`, `inputIsEnterOrSpace`) or the other way around (`const letter`).

src/utils/debounce/index.test.ts
```diff
@@ -0,0 +1,26 @@
+import { debounce } from '.'
+
+jest.useFakeTimers()
+
+describe('debounce()', () => {
+  let fn: jest.Mock
+  let debouncedFunction: Function
```
diogoredin: **suggestion:** As a rule of thumb, I suggest against putting variables inside `describe` blocks. This is because their scope might not be very clear sometimes (Jest can execute tests/ describes in random order). I think placing them at the top of the file is more explicit regarding their scope or saving them in a `this` object in a `beforeEach`.

Having said this, my personal preference goes towards avoiding sharing mocks between tests. It can get hard to track side-effects in large test suites which share the same variables. For instance here, if we were going to share the mock between multiple tests we should reset the mock calls counter between tests (`fn.mockClear`).

src/utils/debounce/index.test.ts
```diff
@@ -0,0 +1,26 @@
+import { debounce } from '.'
+
+jest.useFakeTimers()
```
diogoredin: **suggestion:** I would suggest this to be placed in a `beforeAll` and then we call `jest.useRealTimers();` on a `afterAll` to ensure we don't affect other tests that run after this file.

src/utils/debounce/index.test.ts
```diff
@@ -0,0 +1,26 @@
...
+    debouncedFunction = debounce(fn, 1000)
+  })
+
+  it('execute just once', () => {
+    let i = 0
+
+    while (i < 100) {
+      debouncedFunction()
+      i++
+    }
```
diogoredin: **suggestion:** I think calling it twice explicitly would be enough and could save us some CPU when running this test. What do you think? 🙂

src/components/Select/index.tsx
```diff
@@ -231,6 +247,47 @@ export const Select: React.FC<SelectProps> = ({
...
+
+      const input = query + e.key
+      setQuery(input)
+
+      const index = items.findIndex(item =>
+        item.name.toLowerCase().startsWith(input.toLowerCase())
+      )
+
+      scrollHighlightedItemIntoView()
+      return setHighlightedIndex(index === -1 ? highlightedIndex : index)
```
diogoredin, line 271: **suggestion:** It would be nice to have some tests for these behaviours 😁 You can use the `userEvent` dev package from RTL as we have in Sirius to simulate a click in the select box and then typing in the document (you can query the document). Let me know if you would like to pair-program on this 🙂

diogoredin (dismissed):
Neat feature, this will be really useful 🚀  😀 Left you some suggestions! Let me know what you think 🙂

src/components/Select/index.tsx
bobbonius: Added this ✅

src/components/Select/index.tsx
bobbonius: ✅

src/utils/debounce/index.test.ts
bobbonius: Thanks! ✅

src/utils/debounce/index.test.ts
bobbonius: ✅

src/utils/debounce/index.test.ts
bobbonius: ✅

package.json
```diff
@@ -76,6 +76,7 @@
     "@storybook/storybook-deployer": "^2.8.10",
     "@testing-library/jest-dom": "^5.14.1",
     "@testing-library/react": "^11.2.2",
+    "@testing-library/user-event": "^13.2.1",
```
diogoredin: **nit-pick:** We ended up not using this one. So we can remove it 🧹

diogoredin (commented):
Just a small comment, otherwise LGTM 🚀

diogoredin (approved):
Awesome work 💪  LGTM 🚀

PR #1216, opened 2021-10-13
fix(moneyinput): Safari wouldn't accept decimal numbers

Roohn:
# Description
The default step value in a input where the type is number on safari is `1`
This is why safari wouldn't accept float values, but only integers.

## Changes
- Added step of 0.01 to the moneyInput

## How to test
In safari go to storybook `MoneyInput > with form` and try adding a decimal value and click save.


glenngijsberts (approved):
Nice one 👍

PR #1224, opened 2021-10-18
Update bundler and compiler

### LINKED ISSUE #1081: Improve bundle size & import cost
## Improvement description
Importing from Solar seems to be super expensive in import cost.  Adding or removing multiple components doesn't make a huge difference. 

See:

Screenshot 2021-03-10 at 14 42 49

Screenshot 2021-03-10 at 14 43 54

## Expected behaviour
We should investigate on how we can improve this.

### LINKED ISSUE #1197: Move away from TSDX
## Improvement description
TSDX is no longer maintained so I think it's a good idea to move away from it. Besides that, we want to improve the bundle size so it would be really nice if we can make it smaller and/or investigate how we can properly tree shake it. Now we get some lighthouse warnings that we're using unused code. Not sure if this is related to Sirius or Solar.

## Expected behaviour
Use a different bundler than TSDX, for example Rollup. Already tried to use rollup instead but the bundle seems way bigger. Make sure to test it with Sirius, to see if the performance is improved when you try a new bundler out.

Roohn:
# Description
Solar was using tdsx for bundling and compiling the package. We want move away from this for a couple of reasons:
1. This package was deprecated
2. The tree shaking was not working properly (see screenshots)
3. There wasn't a lot of insight in how it worked (for me personally at least)

## Changes
Using babel for compiling the package and using tsconfig (tsc) for the type declarations
**Note**: We need to use babel for this because emotion has a babel plugin that is required.

## How to test
1. Run `yarn build`
2. Run `npm pack`
3. unzip the package (located in the root foder of solar) and put it in the node_modules of your desired website.
4. Look if the website still loads correctly.

## Screenshots
| Old | New | 
| - | - |
| afbeelding | afbeelding |

## Links
Fixes: https://github.com/TicketSwap/solar/issues/1197
Fixes: https://github.com/TicketSwap/solar/issues/1081


bobbonius:
image

Im getting this error on this component when clicking 'Show dialog'. This doesn't happen on `main`

http://localhost:9001/?path=/story/components-animation-contenttransition--basic

Roohn:
> Im getting this error on this component when clicking 'Show dialog'. This doesn't happen on `main`
> 
> http://localhost:9001/?path=/story/components-animation-contenttransition--basic

Hey good one! It didn't use the babel plugins for storybook. I changed it so now it should be working again 👍🏼

bobbonius (approved):
😎 Very nice

glenngijsberts (approved):
Let's see how it goes 😁

PR #1226, opened 2021-10-21
Theming for storybook

bobbonius:
# Description

I've added theming for Storybook.

## How to test

- Checkout this branch
- `$ yarn && yarn storybook`
- Have a look around and confirm all is still working.

## Screenshots

image

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,9 +1,9 @@
 import React from 'react'
 import { BaseStyles } from '../src/components/BaseStyles'
 import Theme from '../src/components/Theme'
-import { color } from '../src/theme'
 
 export const parameters = {
+  // layout: 'centered',
   backgrounds: {
     default: 'light',
     values: [
--- a/.storybook/theme.js
+++ b/.storybook/theme.js
@@ -1,0 +1,34 @@
+import { create } from '@storybook/theming'
+
+export default create({
+  base: 'light',
+  brandTitle: 'Solar Design System',
+  brandUrl: 'https://ticketswap.com',
+  brandImage:
+    'https://s3-eu-west-1.amazonaws.com/ticketswap/public/downloads/bm-wm-hor-white-blue.png',
+
+  // UI
+  appBg: 'rgba(0, 182, 240, 0.08)',
+  appContentBg: 'white',
+  appBorderColor: 'white',
+  appBorderRadius: 4,
+
+  // Typography
+  fontBase: '"Open Sans", sans-serif',
+  // fontCode: 'monospace',
+
+  // Text colors
+  textColor: 'black',
+  textInverseColor: 'rgba(255,255,255,0.9)',
+
+  // Toolbar default and active colors
+  barTextColor: 'rgba(0, 182, 240, 0.8)',
+  barSelectedColor: 'rgb(0, 182, 240)',
+  barBg: 'rgba(0, 182, 240, 0.25)',
+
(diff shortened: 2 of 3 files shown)


glenngijsberts:
Seems that the prettier test is failing, you can format it with `yarn prettier --write` I think.

.storybook/main.js
```diff
@@ -2,8 +2,11 @@ const path = require('path')
 const toPath = _path => path.join(process.cwd(), _path)
 
 module.exports = {
-  stories: ['../src/**/*.stories.@(ts|js|tsx)'],
-  addons: ['@storybook/addon-essentials'],
+  stories: [
+    '../src/**/*.stories.@(ts|js|tsx)',
+    '../docs/**/*.stories.@(ts|tsx|mdx)',
```
RobVermeer: I guess this is part of another story 😉

.storybook/preview.js
```diff
@@ -17,6 +17,11 @@ export const parameters = {
       },
     ],
   },
+  options: {
+    storySort: {
+      order: ['Guidelines', ['Getting Started', 'Design Tokens']],
```
RobVermeer: This is also for another story right?

package.json
```diff
@@ -78,6 +78,7 @@
     "@emotion/jest": "^11.3.0",
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
+    "@storybook/addon-docs": "^6.3.11",
```
RobVermeer: And this as well

.storybook/preview.js
```diff
@@ -1,9 +1,9 @@
 import React from 'react'
 import { BaseStyles } from '../src/components/BaseStyles'
 import Theme from '../src/components/Theme'
-import { color } from '../src/theme'
 
 export const parameters = {
+  // layout: 'centered',
```
RobVermeer, line 6: The things we don't use we can just remove right?

.storybook/theme.js
```diff
@@ -0,0 +1,34 @@
...
+
+  // UI
+  appBg: 'rgba(0, 182, 240, 0.08)',
+  appContentBg: 'white',
+  appBorderColor: 'white',
+  appBorderRadius: 4,
+
+  // Typography
+  fontBase: '"Open Sans", sans-serif',
+  // fontCode: 'monospace',
```
RobVermeer, line 18: Same for this one

.storybook/theme.js
```diff
@@ -0,0 +1,34 @@
...
+
+export default create({
+  base: 'light',
+  brandTitle: 'Solar Design System',
+  brandUrl: 'https://ticketswap.com',
+  brandImage:
+    'https://s3-eu-west-1.amazonaws.com/ticketswap/public/downloads/bm-wm-hor-white-blue.png',
+
+  // UI
+  appBg: 'rgba(0, 182, 240, 0.08)',
```
RobVermeer, line 11: Would it be an idea to use our own theming colors? So the ones defined at `src/theme.ts`

RobVermeer (approved):
Looks great, really like it! 😍

.storybook/theme.js
bobbonius: @RobVermeer Maybe something for a next iteration. As currently, I can't access the css variable from `theme.ts` in `theme.js`. It's in Solar, but not in Storybook. If that makes sense?

PR #1245, opened 2021-11-05
Improve Select rendering of mobile version

### LINKED ISSUE #1243: Improve Select component rendering
## Improvement description
Right now we are using a `useEffect` to change the type of select box that is rendered (a custom one or a native one for mobile).

The split between one or the other should be done right on the first render, instead of waiting for the component to mount, otherwise, users might see the custom one for a split second (also, we are rendering the custom one without the need on the first render on mobile).

## Expected behaviour
On the first render, we should already see the definitive select (custom or native). We can introduce tests for this and assert the behavior on the first render.

## Additional context
Automatic tests were failing sometimes due to this swapping after the first render. More context can be found [here](https://github.com/TicketSwap/sirius/pull/5190).

diogoredin:
# Description

We were using a `useEffect` to change the type of select box that is rendered (a custom one or a native one for mobile).

The split between one or the other should be done right on the first render, instead of waiting for the component to mount, otherwise, users might see the custom one for a split second (also, we were rendering the custom one without the need on the first render on mobile).

Tests for the rendering of a different component depending on the device used are also introduced.

## Changes

- [x] This has been done
- [ ] I still need to to this

## How to test

- Checkout this branch
- `$ yarn storybook`
- Check the Select component and experiment with different devices
- You can also compile the project and move the output to your own project using Solar and test it there
 
## Screenshots
https://user-images.githubusercontent.com/1096800/140517862-2cec2335-780d-4751-b08b-ed6956d16af3.mov

## To be notified
@TicketSwap/frontend 

## Links
Fixes #1243

### DIFF
--- a/src/components/Select/index.test.tsx
+++ b/src/components/Select/index.test.tsx
@@ -1,6 +1,9 @@
 import React from 'react'
 import { screen, render, fireEvent } from '@testing-library/react'
 import { Select } from '.'
+import * as useDeviceInfo from '../../hooks/useDeviceInfo'
+
+jest.mock('../../hooks/useDeviceInfo')
 
 const languages = [
   { value: 'de', name: 'German' },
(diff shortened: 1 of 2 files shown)


Roohn (approved):
+1 for the tests 👍🏼

RobVermeer (approved):
👍

bobbonius (approved):

PR #1253, opened 2021-11-11
feat(onetimecodeinput): added a OneTimeCodeInput field

bobbonius:
# Description

This component is introduced to handle one time code password to, for example, verify a users
phone number. 

## How to test

- Checkout this branch
- `$ yarn storybook`
- Go to `Inputs/OneTimeCodeInput
- Enter a 5 digit code, manually
- Navigate between the fields from left to right, or right to left
- Use backspace to remove an input and confirm the focus moves to the previous field
- Copy and paste a 5 digit code and paste it into the first field, confirms it spreads the code over all input fields
- Copy and paste a code that is longer than 5 digits and confirm it does nothing at all.
- If you really want to go hardcode: get storybook to run on your mobile browser and ask @glenngijsberts if he wants to send you text that goes says: `Code: 68420`. Check if your phone prompts you with the OneTimeCode from the text message

## Screenshots

image


bobbonius:
😎  Shout out to @diogoredin for helping me with the tests and @RobVermeer for debugging

src/components/OneTimeCodeInput/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+  return (
+    <>
+      {hideLabel ? (
+        <VisuallyHidden>
+          <LabelText>{label}</LabelText>
+        </VisuallyHidden>
+      ) : (
+        <LabelText>{label}</LabelText>
+      )}
```
RobVermeer: Currently the label is not actually a label, is there a reason for that?

src/components/OneTimeCodeInput/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+  label?: string
+  disabled?: boolean
+  helpText?: string
+  onChange?: (value: any) => void
+}
+
+const OneTimeCodeContainer = styled.section<OneTimeCodeContainerProps>`
+  display: grid;
+  grid-template-columns: repeat(${props => props.length}, 1fr);
+  grid-gap: ${space[8]};
```
glenngijsberts, line 29: Nit: Add empty line between the properties and selector

glenngijsberts:
If the first input is not filled in, you can't remove the values of the other inputs.

glenngijsberts:
I can see that we might want to use this input for 2FA authentication as well, and that's always with 6 numbers AFAIK. Might be nice to have a feature where you can input the amount of "boxes" you want to have?

glenngijsberts:
Maybe nice to add a `validate` prop like we have on the Input component?

src/components/OneTimeCodeInput/index.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+export const Basic = () => {
+  return (
+    <Dialog on>
+      {({ getWindowProps }) => (
+        <>
+          <DialogWindow {...getWindowProps()}>
+            <DialogHeader>One Time Code Input</DialogHeader>
+            <DialogBody>
+              <OneTimeCodeInput
+                length={5}
```
glenngijsberts: Since `5` is the default, maybe nice to use `6` instead? Or add another story that shows how you can more or less than 5?

src/components/OneTimeCodeInput/index.tsx
glenngijsberts: Actually, why don't you use the `text-align` in the styled `Input` itself?

src/components/OneTimeCodeInput/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+  ariaLabel: string
+  label?: string
+  disabled?: boolean
+  helpText?: string
+  onChange?: (value: any) => void
+}
+
+const OneTimeCodeContainer = styled.section<OneTimeCodeContainerProps>`
+  display: grid;
+  grid-template-columns: repeat(${props => props.length}, 1fr);
```
glenngijsberts: Thinking out loud: Maybe we should have a min length of 2 and max length of let's say, 8? because now it already looks a bit bad on mobile if you have 8, and even more will overflow the container. I think we can be opinionated about that, by saying that you can only have inputs of those lengths. Otherwise, the design looks really crappy.

src/components/OneTimeCodeInput/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+          <Input
+            onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
+              handleChange(event, index)
+            }}
+            onKeyUp={event => handleKeyUp(event, index)}
+            onFocus={event => event.target.setSelectionRange(1, 1)}
+            value={value}
+            key={index}
+            disabled={disabled}
+            label=""
```
glenngijsberts: Why is this an empty label? 😬

glenngijsberts:
Clicking the label won't select any input. I would assume clicking the label of the input it will select either the first input box, or the box where you were left off.

src/components/OneTimeCodeInput/index.tsx
bobbonius: Because TypeScript complains if there is no label prop used @glenngijsberts :(

src/components/OneTimeCodeInput/index.tsx
bobbonius: @glenngijsberts The `Input` is made with the `fieldStyles` from the `Input` component. If I add it there it won't work because the basic `fieldStyles` will override it and force it left again.

src/components/OneTimeCodeInput/index.tsx
bobbonius: I agree, I think it's should be 6 to be honest

src/components/OneTimeCodeInput/index.tsx
bobbonius: Will fix this

src/components/OneTimeCodeInput/index.tsx
bobbonius: No, will fix this

src/components/OneTimeCodeInput/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+          <LabelText>{label}</LabelText>
+        )}
+      </Label>
+      <OneTimeCodeContainer ref={containerRef} length={length}>
+        {inputValues.map((value, index) => (
+          <Input
+            onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
+              handleChange(event, index)
+            }}
+            onKeyUp={event => handleKeyUp(event, index)}
```
RobVermeer: As discussed, let's change to `keyDown` ⌨️

RobVermeer (approved):
Works very well! Really cool =)

src/components/OneTimeCodeInput/index.tsx
glenngijsberts: @bobbonius ehhhhh, if you declare it like this:

```js
const Input = styled.input`
  ${fieldStyles};
  text-align: center;
`
```

Screenshot 2021-11-16 at 11 40 45

Screenshot 2021-11-16 at 11 41 05

Then the `text-align: center` will overwrite the text-align coming from the `fieldStyles` right? Did you try that?

glenngijsberts (commented):
Everything looks good, but you need to export this component from the `components/index.ts` file, otherwise it won't get bundled :)

glenngijsberts (approved):

PR #1259, opened 2021-11-17
Add icons to Solar

Roohn:
# Description
This will add the icons from Comets into solar

## Changes
- [x] Added all icons
- [x] Changed build process to add `/icons` bundle 
- [x] Added iconography to guidelines in storybook

## How to test
1. run `yarn storybook` to see the new guidelines
2. run `yarn dev-pack`, unzip the package zipfile from `./dist` and place it's content into your own library to see the changes. (Replace the solar package).

## Screenshots
afbeelding


bobbonius (approved):

docs/Guidelines/Iconography/Icons.tsx
```diff
@@ -0,0 +1,299 @@
...
+          <Icon.Copy title="Copy name" size={12} />
+        )}
+      </CopyButton>
+    </IconBlock>
+  )
+}
+
+export const Icons = () => {
+  return (
+    <IconGrid>
```
glenngijsberts, line 77: Can't we make this a bit smarter by mapping over certain values so we don't have to repeat ourselves? 😛  Maybe you can use something like `Object.keys(Icon)` to map over all the icons we import?

Roohn: No I couldn't find a way to do that with the components 😓

package.json
```diff
@@ -20,14 +17,16 @@
...
-    "build:cjs": "BABEL_ENV=cjs babel src --out-dir dist/cjs --extensions '.ts,.tsx' --ignore 'src/**/*.stories.tsx','src/**/*.test.tsx'",
+    "postbuild": "npm run prepdist",
+    "prepdist": "node ./config/prepareDist.js",
     "test": "is-ci \"test:coverage\" \"test:watch\"",
-    "test:coverage": "NODE_ENV=test jest test --coverage",
-    "test:watch": "NODE_ENV=test jest test --watch",
+    "test:coverage": "jest test --coverage",
+    "test:watch": "jest test --watch",
     "lint": "eslint src",
+    "dev-pack": "yarn build && cd dist && npm pack",
```
glenngijsberts, line 29: `pack` is a bit of a weird name 🤔  anyway, can we maybe describe the commands somewhere so we know what each command does and when to use it?

glenngijsberts (approved):
Seems to be working as expected and looks really nice! Can we make sure to deprecate comets? e.g. also archiving it etc. Wondering what the impact will be on performance wise, so let me know when we have a PR to update sirius, so we can test that a bit better.

docs/Guidelines/Iconography/Icons.tsx
glenngijsberts: 😢

package.json
Roohn: I named it like this because the npm command is also `npm pack` 🤔 
used to just make a package file you can use in testing. 

I will add some documentation to the solar package readme on how to test new build etc.

Issue #1060, opened 2021-02-02
Fix HorizontalScroll component


glenngijsberts:
## Improvement description
Currently our HorizontalScroll component is using a lot of properties that are not (longer) supported on every platform/browser. Let's improve that by only using the right properties to create the horizontal scroll effect. 

## Expected behaviour
A refactored HorizontalScroll component.

Example: https://caniuse.com/?search=scroll-snap-type

PR #1260, opened 2021-11-17
Cleanup of `HorizontalScroll` Component

Emacdyz:
# Description
Removed unecessary css from HorizontalScroll Component 

## Changes

I removed unnecessary css that were no longer supported by main browsers anyway. I then tested on safari, firefox and chrome + used browsertsack to try on different mobile devices and browsers.

## How to test
- Make sure the horizontal scroll still works as it should on different browsers 

## Links
#1060


bobbonius (changes requested):
LGTM, in order for the checks to pass you have to make your commit semantic by using `yarn commit`. Let me know if you n eed help with that

PR #1266, opened 2021-11-22
Refactor(horizontalscroll): removal of unecessary/deprecated css

### LINKED ISSUE #1060: Fix HorizontalScroll component

Emacdyz:
# Description

Cleaned up some of the css that were no longer supported by main browsers and
not useful for the horizontal scroll to work as expected

## Links
Fix #1060


bobbonius (approved):
😎

PR #1275, opened 2021-11-30
Export svg wrapper

Roohn:
# Description
Export svgWrapper (formally Icon component) from the icons package.
This component can be used when adding icons outside of Solar while keeping the same style.


glenngijsberts (approved):
😎

PR #1276, opened 2021-11-30
Make Card component with Subtitle and Text the same font-size

diogoredin:
# Description

The Text in the `Card` component currently acts as a kind of sub-subtitle. We want both the `Text` and the `Subtitle` to have the same font-size when they are both defined. There is only one place where this happens in Sirius, which is where we want to define this (the listings inside event-type pages).

## Changes

- [x] This has been done
- [ ] I still need to to this

## How to test

- Checkout this branch
- `$ yarn dev`

## Screenshots
**Before**
Screenshot 2021-11-30 at 18 53 51

**After**
Screenshot 2021-11-30 at 18 53 19

## To be notified
@Marruk 

## Links
[JIRA GIN-1731](https://ticketswap.atlassian.net/browse/GIN-1731)

### DIFF
--- a/src/components/Card/index.stories.tsx
+++ b/src/components/Card/index.stories.tsx
@@ -106,6 +106,44 @@ export const WithText = () => (
 
 WithText.storyName = 'With text'
 
+export const WithTextAndSubtitle = () => (
+  <Grid>
+    <a href="/" style={{ minWidth: 0 }}>
+      <Card
+        title="Eurovision Song Contest - 1st Semi Final Jury Show"
+        subtitle="I'm a subtitle"
+        text="Lorem ipsum dolor sit amet."
+        verticalAlign={CardVerticalAlign.top}
+      />
+    </a>
+    <a href="/" style={{ minWidth: 0 }}>
+      <Card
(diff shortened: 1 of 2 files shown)


Roohn (approved):

Marruk:
All sounds a bit hacky to me, but if the frontenders are happy I'm happy 🙂

When I have some time I can take a look where we actually use that smaller font size with the subtitle, maybe we can just make all of them large and we can clean this component up a bit.

diogoredin:
> All sounds a bit hacky to me, but if the frontenders are happy I'm happy 🙂
> 
> When I have some time I can take a look where we actually use that smaller font size with the subtitle, maybe we can just make all of them large and we can clean this component up a bit.

@Marruk I also don't like it 😅 There are far too many use cases for this component, which actually require different sizes and behaviours in different situations. I started collecting the use cases here: https://www.notion.so/ticketswap/Card-Component-Usages-b27b25a8ada84921a9b5fbcc8345d901 there are probably more, though. Would be cool maybe to have these different cards (or variants here).

Marruk:
Screenshot 2021-12-02 at 10 25 01

PR #1283, opened 2021-12-06
feat(select): make it possible to add onClick actions to select items

glenngijsberts:
# Description

This change makes it possible to add an option to the select items that is not used as item but more
as action after selecting it. By using the `type` and `onClick` attribute it's possible to run a
function after selecting an item (without actually selecting the item).

## How to test

- Checkout
- See `withActions` story
- Verify everything is working as expected
- Verify mobile view is also working as expected

## Screenshots
Screenshot 2021-12-13 at 11 55 14

### DIFF
--- a/src/components/Select/index.tsx
+++ b/src/components/Select/index.tsx
@@ -37,12 +37,19 @@ const Container = styled.div`
 
 Container.defaultProps = { className: 'select' }
 
+export enum SelectItemType {
+  select = 'select',
+  action = 'action',
+}
+
 type SelectItem = {
   value: string
   name: string
   displayName?: string
   leftAdornment?: ReactNode
   rightAdornment?: ReactNode
+  type?: SelectItemType
+  onClick?: Function
 }
 
 export interface SelectProps extends Omit<InputProps, 'onChange'> {
@@ -136,6 +143,7 @@ export const Select: FC<SelectProps> = ({
   const containerRef = useRef<HTMLDivElement>(null)
   const inputRef = useRef<HTMLInputElement>(null)
   const menuRef = useRef<HTMLUListElement>(null)
+  const mobileSelectRef = useRef<HTMLSelectElement>(null)
   let itemRefs: HTMLLIElement[] = []
   const labelId = `${id}-label`
   const menuId = `${id}-menu`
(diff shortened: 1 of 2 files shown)


Roohn:
When trying to select the options with the onClick through keyboard actions (arrow down and clicking enter) it won't fire the onClick :)

src/components/Select/index.tsx
```diff
@@ -37,12 +37,19 @@ const Container = styled.div`
...
+}
+
 type SelectItem = {
   value: string
   name: string
   displayName?: string
   leftAdornment?: ReactNode
   rightAdornment?: ReactNode
+  type?: SelectItemType
+  onClick?: Function
```
Roohn, line 52: You can add something called [`Discriminated unions`](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions) to make sure there is always an onClick when the type is `action`.

Something like this:
```js
type CommonSelectItem = {
  value: string
  name: string
  displayName?: string
  leftAdornment?: ReactNode
  rightAdornment?: ReactNode
}

type TypeProps = { type: 'select'; onClick?: never } | { type: 'action'; onClick: Function }

type SelectItem = CommonSelectItem & TypeProps
```

bobbonius:
I think it's nice to have something like this:

image

Not entirely sure it should look like this but I do think it's important to distinguish the difference of an option or an action

glenngijsberts:
> I think it's nice to have something like this:
> 
> image
> 
> Not entirely sure it should look like this but I do think it's important to distinguish the difference of an option or an action

Nice idea. I've added it to the example.

Since the items are declared outside of the component I haven't added this to the component itself. This will give more control to the user that's using the component as well.

glenngijsberts:
> When trying to select the options with the onClick through keyboard actions (arrow down and clicking enter) it won't fire the onClick :)

@Roohn fixed!

src/components/Select/index.tsx
glenngijsberts: Can't get this to work, and I can't find a way to default the `type` prop on a `SelectItem`, which makes this a bit annoying because then you need to declare for every item the type, although most of the items are always `select`.

Roohn (approved):

PR #1291, opened 2021-12-13
Introduce DatePicker Component

diogoredin:
# Description

This PR introduces a new **DatePicker** component. Two instances of this date picker will replace the existing date range picker used in TicketSwap. A clear button will be implemented in a future PR.

The **DatePicker** supports the selection of dates in the future, past, or both. The native i18n functions for dates are used, to save on bundle size and to not force users to use a specific time library.

## Screenshots
dialog

## Changes

- [x] This has been done
- [ ] I still need to to this

## How to test

- Checkout this branch
- `$ yarn dev`

## To be notified
@Marruk @TicketSwap/frontend 

## Links
This will replace the existing range DatePicker used in TicketSwap, more details [here](https://www.figma.com/file/a4Lq5vWjnJrHAccauMLBR7/Ongoing-Events?node-id=111%3A13).

### DIFF
--- a/src/components/DatePicker/CalendarDialog.tsx
+++ b/src/components/DatePicker/CalendarDialog.tsx
@@ -1,0 +1,177 @@
+import React from 'react'
+import { useState } from 'react'
+import styled from '@emotion/styled'
+import {
+  Dialog,
+  DialogAdornment,
+  DialogBody,
+  DialogHeader,
+  DialogWindow,
+} from '../Dialog'
+import { Select } from '../Select'
+import { Text } from '../Text'
+import { H6 } from '../Heading'
+import { CloseAlt } from '../../icons'
+import { space, color, device } from '../../theme'
+import Day from './Day'
+import {
+  TimeFrame,
+  isSameDate,
+  getWeekdays,
+  getMonths,
+  getYears,
+  isPastDate,
+  isFutureDate,
+} from '../../utils/dates'
+
+const Selects = styled.div`
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  grid-gap: ${space[8]};
+`
+
+const CalendarGrid = styled.div`
+  display: grid;
+  grid-template-columns: repeat(7, 1fr);
+  grid-gap: ${space[4]};
+`
+
+const CalendarTitles = styled.div`
+  display: grid;
+  grid-template-columns: repeat(7, 1fr);
+  grid-gap: ${space[4]};
+  margin-top: ${space[24]};
+  margin-bottom: ${space[8]};
+`
+
+const CalendarTitle = styled(H6)`
+  text-align: center;
+`
+
+const InfoText = styled(Text)`
+  margin-top: ${space[16]};
+  color: ${color.spaceMedium};
+`
+
+const StyledDialogBody = styled(DialogBody)`
+  max-height: 70vh;
+
+  @media ${device.mobileL} {
+    max-height: none;
+  }
+`
+
+interface CalendarDialogProps {
+  date: Date | null
+  title?: string
+  timeFrame?: TimeFrame
+  monthLabel: string
+  yearLabel: string
+  info?: string
+  isOpen: boolean
+  locale: string
+  close: () => void
+  onChange: (date: Date) => void
+}
+
+const CalendarDialog = ({
+  date,
+  title,
+  timeFrame,
+  monthLabel,
+  yearLabel,
+  info,
+  isOpen,
+  close,
+  locale,
+  onChange,
+}: CalendarDialogProps) => {
+  const month = date?.getMonth() || new Date().getMonth()
+  const year = date?.getFullYear() || new Date().getFullYear()
+
+  const [selectedMonth, setSelectedMonth] = useState(month)
+  const [selectedYear, setSelectedYear] = useState(year)
+
+  const daysInMonth = new Date(selectedYear, selectedMonth + 1, 0).getDate()
+  const firstDayOfMonth = new Date(selectedYear, selectedMonth, 1).getDay()
+
+  const weekdays = getWeekdays(locale)
+  const months = getMonths(locale)
+  const years = getYears(timeFrame)
+
+  const days: Array<React.ReactNode> = []
+
+  for (let i = 1; i < daysInMonth + 1; i++) {
+    const day = new Date(selectedYear, selectedMonth, i)
+    const isDisabled =
+      (timeFrame === TimeFrame.future && isPastDate(day)) ||
+      (timeFrame === TimeFrame.past && isFutureDate(day))
+
+    days.push(
+      <Day
+        key={i}
+        date={day}
+        isDisabled={isDisabled}
+        isSelected={!!date && isSameDate(date, day)}
+        onSelect={onChange}
+      />
+    )
+  }
+
+  for (let i = 1; i < firstDayOfMonth; i++) {
+    days.unshift(<div key={`prev-${i}`} />)
+  }
+
+  return (
+    <Dialog on={isOpen}>
+      {({ getWindowProps }) => (
+        <DialogWindow
+          {...getWindowProps({
+            onClick: close,
+            onEscKeyDown: close,
+          })}
+        >
+          <DialogHeader>
+            {title}
+            <DialogAdornment right>
+              <button onClick={close}>
+                <CloseAlt size={24} />
+              </button>
+            </DialogAdornment>
+          </DialogHeader>
+          <StyledDialogBody>
+            <Selects>
+              <Select
+                id="month"
+                items={months}
+                label={monthLabel}
+                initialSelectedItem={months.find(
+                  i => i.value === String(month)
+                )}
+                hideLabel
+                onChange={selected => setSelectedMonth(Number(selected.value))}
+              />
+              <Select
+                id="year"
+                items={years}
+                label={yearLabel}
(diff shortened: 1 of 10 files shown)


src/components/DatePicker/CalendarDialog.tsx
```diff
@@ -0,0 +1,169 @@
...
+const Selects = styled.div`
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  grid-gap: ${space[8]};
+`
+
+const CalendarGrid = styled.div`
+  display: grid;
+  grid-template-columns: repeat(7, 1fr);
+  grid-gap: 4px;
```
Roohn:
```suggestion
  grid-gap: ${space[4]};
```

src/components/DatePicker/Day.tsx
```diff
@@ -0,0 +1,88 @@
...
+    if (!isDisabled) {
+      onSelect(date)
+    }
+  }
+
+  return (
+    <DayButton
+      selected={isSelected}
+      aria-selected={isSelected}
+      {...(isDisabled ? { disabled: true } : {})}
```
Roohn, line 79: This can just be a boolean with either true or false right? Otherwise the type should be `disabled?: true` instead of `disabled?: boolean` 🤔

src/utils/dates/index.ts
```diff
@@ -0,0 +1,86 @@
+export enum TimeFrame {
+  all = 'all',
```
Roohn, line 2: Isn't it easier to call this `present` or `everything`? Not sure what all means 🤔

src/components/DatePicker/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+          close()
+        }}
+      />
+      <StyledButton onClick={toggle}>
+        <Calendar size={24} color={color.spaceMedium} />
+        <ButtonText hasDate={!!date}>
+          {!!date
+            ? `${date.getDate()} ${date.toLocaleString(locale, {
+                month: 'short',
+              })} ${date.getFullYear()}`
```
Roohn: This is not localized. I don't know the specifics but not all countries use `day, month, year`. Also not when writing it in full like this

Roohn:
When resizing to 420px width it will all look a bit aligned to the left (see guidelines in screenshot)
afbeelding

src/components/DatePicker/Day.tsx
diogoredin: Actually, this type is not needed since it will inherit the prop from the HTML button type. thanks for noticing  🙏

src/utils/dates/index.ts
diogoredin: `all` means past and future in this context. Not sure if `everything` will convey more meaning, though. I also thought of `broad`, wdyt?

src/components/DatePicker/index.tsx
diogoredin: Nice catch, that's very true! We can actually do:

```
new Date().toLocaleString(locale, {
  day: 'numeric',
  month: 'short',
  year: 'numeric',
})
```

`Dec 13, 2021`

However, it adds a comma, which is not according to our new guidelines. I will check with @Marruk what we can do without adding any libs here 😅 

Another solution I thought of is leaving the responsibility of displaying the value to the user of the component, allowing the usage of any lib outside, but I'm not sure if I like that yet 💭

src/components/DatePicker/CalendarDialog.tsx
```diff
@@ -0,0 +1,169 @@
...
+        >
+          <DialogHeader>
+            {title}
+            <DialogAdornment right>
+              <button onClick={close}>
+                <CloseAlt size={24} />
+              </button>
+            </DialogAdornment>
+          </DialogHeader>
+          <DialogBody>
```
bobbonius: Maybe it's nice to style this DialogBody a bit so that on mobile the calendar doesn't move around. I think changing the `max-height` to let's say `70vh` should probably work

src/components/DatePicker/CalendarDialog.tsx
```diff
@@ -0,0 +1,169 @@
...
+                initialSelectedItem={years.find(i => i.value === String(year))}
+                hideLabel
+                onChange={selected => setSelectedYear(Number(selected.value))}
+              />
+            </Selects>
+            <CalendarTitles>
+              {weekdays.map((weekday, index) => (
+                <CalendarTitle key={`day-${index}`}>{weekday}</CalendarTitle>
+              ))}
+            </CalendarTitles>
```
bobbonius, line 167: Not sure if this is because Storybook's viewport changer because I only have this if I use there option to change to viewport to 'small mobile' but when I do it looks like this:

image

bobbonius (commented):
I think it would be nice to select a date range. All though I don't know if it's something we necessarily, I think it should definitly be part of a date picker. So for example: using shift to select a range from day 1 to 5

src/components/DatePicker/index.tsx
bobbonius: @diogoredin Why not use `new Intl.DateTimeFormat('en-US').format(date))`

[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat)

diogoredin: It's also an option, but it has the same issue as the `toLocaleString` as it adds chars that do not conform with our new guidelines 😢  I'm checking this with @Marruk.

Screenshot 2021-12-14 at 16 30 38

src/components/DatePicker/CalendarDialog.tsx
diogoredin: Can explain what is moving? 😄  The Body already has a maximum height of `50vh` on mobile.

src/components/DatePicker/CalendarDialog.tsx
diogoredin: Nice catch! This was not okay on mobile. Should be fixed now, let me know what you think. 🙂 

Screenshot 2021-12-14 at 16 29 20

Screenshot 2021-12-14 at 16 28 58

diogoredin:
> I think it would be nice to select a date range. All though I don't know if it's something we necessarily, I think it should definitly be part of a date picker. So for example: using shift to select a range from day 1 to 5

To give a bit more context, the date picker we currently have on TicketSwap supports date ranges, however, we are going to replace all those instances with two single date pickers instead (the one we are introducing now). This is because the UX was not the best, it was hard to select a range covering multiple months or years. @Marruk can maybe add more reasons for this decision 😄 

We might still add back the functionality of supporting ranges in the DatePicker, but for now, the requirement is only for selecting a single date.

diogoredin:
> When resizing to 420px width it will all look a bit aligned to the left (see guidelines in screenshot)

Should be okay now! 🙂

Marruk:
> @Marruk can maybe add more reasons for this decision 😄

The reason I opted for a single date date picker instead of a date range one is because selecting a single date with the date range picker was quite awkward in some scenarios. I expect that use case to be at least as common as people selecting a date range. The component can still have a date range version of course but we won't be using those for now.

As for the dates: let's go with `new Date().toLocaleString(locale, { day: 'numeric', month: 'short', year: 'numeric' })` for now. I have a ticket on my backlog to look (and standardise) all displayed dates across all our clients so I'll make sure to give internationalisation more thought there.

src/components/DatePicker/CalendarDialog.tsx
bobbonius: @diogoredin See video below. I have to say, this is what happens in the viewport on storybook on smallest screen. Might behave completely different on an actual mobile device. Havent tested that..

https://user-images.githubusercontent.com/12168698/146080712-e59d0950-90f7-4b6b-b118-9f8d1387ac2b.mov

bobbonius: Btw, after the part where I chang it to 70vh, i try to move it again but that doesnt really show 😂

src/components/DatePicker/CalendarDialog.tsx
bobbonius: Still happens with me

src/utils/dates/index.ts
Roohn: True, I have no beter naming suggestion so this is fine 😄

Roohn:
There is still a weird jump between `768px` width and the mobile view. See the right side of the screenshot
afbeelding

src/components/DatePicker/CalendarDialog.tsx
diogoredin: Got it, thanks! Increased this for small screens.

src/components/DatePicker/CalendarDialog.tsx
diogoredin: This is really weird, maybe we can make a call to check this? I tested on my phone and all mobile phones on Chrome Dev Tools and seems okay.

https://user-images.githubusercontent.com/1096800/146166741-6818dc3e-c51c-4d68-8ccb-99c7dc264461.MP4

diogoredin:
@Roohn @bobbonius third time is the charm? 😛 Adjusted the grid (it needed smaller spacing for smaller screens indeed) and tested on all mobile devices available on the dev tools. Also tested with resizing the window. Should be okay now, but if not ping me 🙏

bobbonius (approved):
😎 Looking good!!

Roohn (approved):
📅 📅 👈🏼

PR #1302, opened 2021-12-23
Make DatePicker timezone aware

diogoredin:
# Description

As a user of this component, I might want to display the days in a different time zone. For instance, when I'm picking an event date in Australia, the day of the event might be on a different day / week day than my local time. 

## Changes

- [ ] This has been done
- [x] I still need to to this

## How to test

- Checkout this branch
- `$ yarn dev`

## To be notified

@Marruk @samybasset


Marruk:
@diogoredin what happened?

PR #1316, opened 2022-01-05
feat (DatePicker): Custom time frame with date range [GIN-1846]

Emacdyz:
# Description
After introducing the new day picker, we want to facilitate the UX for ongoing events by allowing the user to pick dates only from the dateRange of the event. The rest of the dates will be disabled + implementation of clear button (only showing when there is a value in the date field). 

## Changes

- [x] This has been done
- [ ] I still need to to this

## How to test

- Checkout this branch
- `$ yarn storybook` 
- Go to datePicker story: ' With time range '
- Check that it disabled the dates outside of the dateRange pass to the component 
- Check that both the year and month selects are limited to the months and year within the validity 
- Check that the datePicker opens on the first month of the start date of the dateRange
- Go to the story ' With reset button '
- Make sure it clear both dates on click x and that the x shows only if the date value is not null

## Screenshots
Screenshot 2022-01-05 at 17 49 27
Screenshot 2022-01-10 at 09 44 18

## To be notified

## Links
https://ticketswap.atlassian.net/browse/GIN-1846

### DIFF
--- a/src/components/DatePicker/CalendarDialog.tsx
+++ b/src/components/DatePicker/CalendarDialog.tsx
@@ -117,7 +117,11 @@ const CalendarDialog = ({
   const days: Array<React.ReactNode> = []
 
   useEffect(() => {
-    if (!date) return
+    if (!date) {
+      setSelectedMonth(initialMonth)
+      setSelectedYear(initialYear)
+      return
+    }
 
     const month = date.getMonth()
     const year = date.getFullYear()
--- a/src/components/DatePicker/CalendarDialog.tsx
+++ b/src/components/DatePicker/CalendarDialog.tsx
@@ -1,5 +1,4 @@
-import React, { useEffect } from 'react'
-import { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import styled from '@emotion/styled'
 import {
   Dialog,
@@ -65,6 +64,7 @@ interface CalendarDialogProps {
   date: Date | null
   title?: string
   timeFrame?: TimeFrame
+  dateRange?: { start: Date; end: Date }
   monthLabel: string
   yearLabel: string
   info?: string
@@ -78,6 +78,7 @@ const CalendarDialog = ({
   date,
   title,
   timeFrame,
+  dateRange,
   monthLabel,
   yearLabel,
   info,
@@ -86,19 +87,33 @@ const CalendarDialog = ({
   locale,
   onChange,
 }: CalendarDialogProps) => {
-  const month = date ? date.getMonth() : new Date().getMonth()
-  const year = date ? date.getFullYear() : new Date().getFullYear()
+  const dateRangeMonth = dateRange
+    ? dateRange.start.getMonth()
+    : new Date().getMonth()
+  const dateRangeYear = dateRange
+    ? dateRange.start.getFullYear()
+    : new Date().getFullYear()
 
-  const [selectedMonth, setSelectedMonth] = useState(month)
-  const [selectedYear, setSelectedYear] = useState(year)
+  const initialMonth = date ? date.getMonth() : dateRangeMonth
+  const initialYear = date ? date.getFullYear() : dateRangeYear
+
+  const [selectedMonth, setSelectedMonth] = useState(initialMonth)
+  const [selectedYear, setSelectedYear] = useState(initialYear)
 
   const daysInMonth = new Date(selectedYear, selectedMonth + 1, 0).getDate()
   const firstDayOfMonth = new Date(selectedYear, selectedMonth, 1).getDay()
 
   const weekdays = getWeekdays(locale)
-  const months = getMonths(locale)
-  const years = getYears(timeFrame)
-
+  const months = getMonths(locale, dateRange)
+  const years = getYears({
+    timeFrame: timeFrame || TimeFrame.future,
+    range: dateRange
+      ? dateRange.end.getFullYear() - dateRange.start.getFullYear()
+      : 10,
+    yearOfReference: dateRange
+      ? dateRange.start.getFullYear()
+      : new Date().getFullYear(),
+  })
   const days: Array<React.ReactNode> = []
 
   useEffect(() => {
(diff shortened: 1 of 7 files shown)


src/components/DatePicker/CalendarDialog.tsx
```diff
@@ -86,19 +87,38 @@ const CalendarDialog = ({
...
-  const years = getYears(timeFrame)
-
+  const allMonths = getMonths(locale)
+  const months = dateRange
+    ? allMonths.filter(
+        month =>
+          Number(month.value) >= dateRange.start.getMonth() &&
+          Number(month.value) <= dateRange.end.getMonth()
+      )
+    : allMonths
```
diogoredin: **suggestion:** For years we do the filtering within the utility. What do you think about doing the same here, to be consistent in the calls to the utils?

src/components/DatePicker/CalendarDialog.tsx
```diff
@@ -86,19 +87,38 @@ const CalendarDialog = ({
...
+  const years = dateRange
+    ? getYears(
+        timeFrame,
+        dateRange.end.getFullYear() - dateRange.start.getFullYear(),
+        dateRange.start.getFullYear()
```
diogoredin: **suggestion:** What do you think about having the util receive an object with named properties to make it easier to understand what the second and last arguments do?

src/components/DatePicker/CalendarDialog.tsx
```diff
@@ -65,6 +64,7 @@ interface CalendarDialogProps {
   date: Date | null
   title?: string
   timeFrame?: TimeFrame
+  dateRange?: { start: Date; end: Date }
```
diogoredin, line 67: **suggestion:** What do you think about making this required if the timeframe is custom?

Something like this:

```ts
interface DatePickerProps {
  date?: Date | null
  placeholder: string
  title?: string
  info?: string
  monthLabel: string
  yearLabel: string
  locale?: string
  onChange: (date: Date) => void
}

type TimeFrameProps =
  | { timeFrame?: never; dateRange?: never }
  | { timeFrame?: TimeFrame.all; dateRange?: never }
  | { timeFrame?: TimeFrame.past; dateRange?: never }
  | { timeFrame?: TimeFrame.future; dateRange?: never }
  | { timeFrame?: TimeFrame.custom; dateRange: { start: Date; end: Date } }

...

export const DatePicker = ({
  date = null,
  placeholder,
  title,
  timeFrame = TimeFrame.future,
  dateRange,
  info,
  monthLabel,
  yearLabel,
  locale = 'en-US',
  onChange,
}: DatePickerProps & TimeFrameProps)
```

src/components/DatePicker/CalendarDialog.tsx
```diff
@@ -156,7 +179,7 @@ const CalendarDialog = ({
                 items={months}
                 label={monthLabel}
                 initialSelectedItem={months.find(
-                  i => i.value === String(month)
+                  i => i.value === String(selectedMonth)
```
diogoredin: **question:** Not sure if we want to change the initial data (because the selected can change but `initialMonth` and `initialYear` not. What do you think?

diogoredin (commented):
Awesome work @Emacdyz 🙌  Overall LGTM! What do you think about also adding some tests for the new scenarios?

src/components/DatePicker/CalendarDialog.tsx
diogoredin: For some reason, Typescript doesn't recognize the enums as being part of TimeFrame. Let's move on without this for now, and revisit later.

src/components/DatePicker/index.tsx
```diff
@@ -88,16 +106,30 @@ export const DatePicker = ({
...
+          </ButtonText>
+        </div>
+        {onReset && date && (
+          <ResetButton
+            onClick={e => {
+              onReset()
+              e.stopPropagation()
+            }}
+            type="button"
+            data-testid="reset-button"
```
diogoredin: **suggestion (if-minor):** We could add an `aria-label` here instead so screen readers pick up on this. And then on the tests, we can use that instead of the `data-testid`. What do you think? The only inconvenience is that the aria-label needs to be an argument since it needs to come translated from the user of the component.

src/components/DatePicker/index.stories.tsx
```diff
@@ -87,6 +87,27 @@ export const All = () => {
 
 All.storyName = 'All dates'
 
+export const WithTimeRange = () => {
```
diogoredin, line 90: **suggestion (if-minor):** On `src/components/DatePicker/CalendarDialog.test.tsx` line 386 we have tests for the all, past, and future variants. Would be cool to have the same for the new 'custom' variant. What do you think?

src/components/DatePicker/index.stories.tsx
```diff
@@ -104,6 +125,24 @@ export const WithTitle = () => {
 
 WithTitle.storyName = 'With title'
 
+export const WithResetButton = () => {
```
diogoredin, line 128: **suggestion (if-minor):** Would be cool to have a test for this as well. e.g. provide an initial date, search for the reset button and click it, check the date is no longer in the screen.

diogoredin (approved):
Awesome work, LGTM 🚀 Left just some remarks if you have time to add some more tests.

src/components/DatePicker/index.tsx
Emacdyz: I'm not sure I see the point, and already in input component the reset-button is also using `data-testid` so I thought for the sake of consistency, i would keep it like that

diogoredin: Strong opinion ahead 😅 Forcing ourselves to use accessibility getters for testing ensures we are improving accessibility any time we test something. Besides this by using an accessibility getter instead of `data-testid` we are not tying the tests to implementation details.

Emacdyz: Fixed!

Roohn:
I get this in my storybook, only for the story that this PR adds
afbeelding

bobbonius (approved):
😎 LGTM

Roohn (approved):
Firefox works now too 👍🏼 👍🏼

PR #1337, opened 2022-01-26
Fix the auto import of solar components in IDE's

Roohn:
# Description
The declaration files of .ts files were not created because they weren't included in the tsconfig.
This caused your IDE to import the components/utils not from the index (@ticketswap/solar) but from the specific file they are in (@ticketswap/solar/button). This is fixed now 🎉 

## Changes
Added `.ts` files to the tsconfig.


glenngijsberts (approved):
O YEAH

PR #1345, opened 2022-02-02
Add new color palette

Roohn:
# Description
This PR will introduce our new color palette & semantic color system. It will remove & replace 90% of our current color palette, so this includes breaking changes.

The following colors will be removed:

```
  space: 'var(--space)',
  spaceDark: 'var(--spaceDark)',
  spaceMedium: 'var(--spaceMedium)',
  spaceMediumAlpha: 'var(--spaceMediumAlpha)',
  spaceLight: 'var(--spaceLight)',
  spaceLightAlpha: 'var(--spaceLightAlpha)',
  spaceLighter: 'var(--spaceLighter)',
  spaceLighterAlpha: 'var(--spaceLighterAlpha)',
  spaceLightest: 'var(--spaceLightest)',
  spaceLightestAlpha: 'var(--spaceLightestAlpha)',
  nova: 'var(--nova)',
  novaLight: 'var(--novaLight)',
  novaLightAlpha: 'var(--novaLightAlpha)',
  novaLighterAlpha: 'var(--novaLighterAlpha)',
  novaLightestAlpha: 'var(--novaLightestAlpha)',
  stardust: 'var(--stardust)',
  stardustLight: 'var(--stardustLight)',
  stardustLightAlpha: 'var(--stardustLightAlpha)',
  stardustLighter: 'var(--stardustLighter)',
  stardustLighterAlpha: 'var(--stardustLighterAlpha)',
  earthLight: 'var(--earthLight)',
  earthLightAlpha: 'var(--earthLightAlpha)',
  earthLighter: 'var(--earthLighter)',
  earthLighterAlpha: 'var(--earthLighterAlpha)',
  earthLightest: 'var(--earthLightest)',
  earthLightestAlpha: 'var(--earthLightestAlpha)',
  sky: 'var(--sky)',
  skyAlpha: 'var(--skyAlpha)',
  skyLight: 'var(--skyLight)',
  skyLightAlpha: 'var(--skyLightAlpha)',
  titanLight: 'var(--titanLight)',
  titanLightAlpha: 'var(--titanLightAlpha)',
  titanLighter: 'var(--titanLighter)',
  titanLighterAlpha: 'var(--titanLighterAlpha)',
  titanLightest: 'var(--titanLightest)',
  titanLightestAlpha: 'var(--titanLightestAlpha)',
  mars: 'var(--mars)',
  marsLight: 'var(--marsLight)',
  marsLightAlpha: 'var(--marsLightAlpha)',
  marsLighter: 'var(--marsLighter)',
  marsLighterAlpha: 'var(--marsLighterAlpha)',
  marsLightest: 'var(--marsLightest)',
  marsLightestAlpha: 'var(--marsLightestAlpha)',
  sunLight: 'var(--sunLight)',
  sunLightAlpha: 'var(--sunLightAlpha)',
  sunLighter: 'var(--sunLighter)',
  sunLighterAlpha: 'var(--sunLighterAlpha)',
  sunLightest: 'var(--sunLightest)',
  sunLightestAlpha: 'var(--sunLightestAlpha)',
  facebook: 'var(--facebook)',
  facebookLight: 'var(--facebookLight)',
  twitter: 'var(--twitter)',
  twitterLight: 'var(--twitterLight)',
  whatsapp: 'var(--whatsapp)',
```

Check http://localhost:9001/?path=/story/guidelines-colors--page for new colors.

## How to test
- Checkout this branch
- `$ yarn dev`
- `$ yarn storybook`

## To be notified

@Marruk


src/global-styles.ts
```diff
@@ -215,14 +251,15 @@ export const globalStyles = css`
     margin: 0;
     border: none;
     font: inherit;
-    color: ${color.space};
+    color: ${color.foreground};
```
Marruk: Any reason why this is not `action`? Seeing that it was `space` before, but since it's a button doesn't it make sense to be blue?

glenngijsberts: The default of a the html button is also black. Why should we change it to blue? Our default button would look like a link, which it isn't, and if you need that then you should use a link probably.

Marruk: The default of a html button has a background and border that makes it look like a button. That is not the case here, which means that if someone only uses the button element (without any classes) it wouldn't look like something clickable.

¯\_(ツ)_/¯

Roohn: We mostly use it for wrapping an icon. An info icon that opens a dialog for example. These are mostly black, so we can change this to be blue but than we need to keep these things in mind 🤔 
It is not a bad idea to force you to make clickable things a `action` color, but this would change the designs ofcourse

Marruk: Ah makes sense. Wouldn't do any unnecessary changes like that here since this project is big enough, maybe we can create a new Github Issue to look into this at some point instead.

bobbonius (approved):
😎 Nice work, I think it's a great improvement!

PR #1346, opened 2022-02-03
Fix ContentDialogHeader title color in dark mode

glenngijsberts:
# Description

We have a style bug in our production app, where the titles of the ContentDialogHeader (e.g. discovery feed) are using colors from light mode. This PR will make sure we use proper colors for dark mode.

## How to test

It's a bit hard to test, because dialogs in Storybook are not properly using dark mode due to the fact that the portal of dialogs are outside the storybook portal. What you can do is:

* Use element inspector
* Inspect the dialog
* Add `data-theme="dark"` to a parent div
* Verify the dialog is now in dark mode, with proper colors

## Screenshots

New
Screenshot 2022-02-03 at 14 16 59

Old
Screenshot 2022-02-03 at 14 17 26

### DIFF
--- a/src/components/ContentDialog/index.tsx
+++ b/src/components/ContentDialog/index.tsx
@@ -282,6 +282,18 @@ const StyledClose = styled(Close)`
 
 const Subtitle = styled(Text)`
   color: ${color.stardustLightAlpha};
+
+  [data-theme='dark'] & {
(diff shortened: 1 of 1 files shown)


diogoredin (approved):
LGTM 🚀

Roohn (approved):

PR #1353, opened 2022-02-08
feat(datePicker): allowing to hide frontAdornment

Emacdyz:
# Description

Along with the revamp of the TicketAlert modals, it was required to remove the calendar adornment in the front. This PR is hiding it if the props is false. `withFrontAdornment` default to `true`

## Changes

- [x] This has been done
- [ ] I still need to to this

## How to test

- Checkout this branch
- `$ yarn run storybook`
- search for datePicker 
- check that there is no calendar icon in the front in the story "With no front adornment"
- verify there is a calendar icon in the other stories 

## Screenshots
Screenshot 2022-02-08 at 16 17 43
Screenshot 2022-02-08 at 16 17 52

## To be notified
@TicketSwap/frontend 

## Links
https://ticketswap.atlassian.net/browse/GIN-1787

### DIFF
--- a/src/components/DatePicker/index.tsx
+++ b/src/components/DatePicker/index.tsx
@@ -1,14 +1,15 @@
 import React from 'react'
-import { useState } from 'react'
+import { useState, ReactNode } from 'react'
 import styled from '@emotion/styled'
 import { Text } from '../Text'
-import { Calendar, CloseRounded } from '../../icons'
+import { CloseRounded } from '../../icons'
 import { space, radius, color, device, transition } from '../../theme'
 import CalendarDialog from './CalendarDialog'
 import { TimeFrame } from '../../utils/dates'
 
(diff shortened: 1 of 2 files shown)


diogoredin (approved):
Now that I see this, maybe we can pass the adornment as a prop? So we can also change it, as we do in other components, what do you think? Approving because this approach is fine as well 🚀

src/components/DatePicker/index.tsx
```diff
@@ -83,6 +85,7 @@ export const DatePicker = ({
   resetLabel,
   onChange,
   onReset,
+  withFrontAdornment = true,
```
Roohn: Could you rename it to `withLeftAdornment`? To be consistent with all other adornments?

Roohn (approved):

PR #1356, opened 2022-02-08
Replace Semantic Pull Request app for action

Roohn:
## Desciption

This PR will replace the Semantic Pull Request app with a github action called `semantic_commits`

It basically does the same but we have more control over it and it doesn't break every week.

afbeelding

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,16 @@ jobs:
       - run: yarn install
       - run: yarn validate
 
+  semantic_commits:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -24,6 +24,14 @@ jobs:
       - run: yarn install
       - run: yarn validate
 
+  semantic_commits:
+    runs-on: ubuntu-latest
```
glenngijsberts, line 28: Maybe nice to give it a name like the other jobs.

glenngijsberts (approved):
I would say that we still don't have a lot of control using someones else action but I guess this is fine for now.

Roohn:
> I would say that we still don't have a lot of control using someones else action but I guess this is fine for now.

I mean that you can supply a config file and have some contol over it that way 😄

PR #1369, opened 2022-02-16
Toast: don't notify on unmount

Roohn:
# Description
Don't run the notify funtion when the toastProvider is not mounted.
This should help with errors about updating state on unmounted components when testing using the toast component

## Changes
Added a check for mount when calling the notify function of the toast.


src/components/Toast/index.test.tsx
```diff
@@ -42,9 +44,9 @@ describe('Toast', () => {
...
       </ToastProvider>
     )
-    fireEvent.click(getByText(/show toast/i))
-    expect(getByText(/payment failed/i)).toBeInTheDocument()
-    fireEvent.click(getByText(/discard/i))
-    expect(queryByText(/payment failed/i)).toBeNull()
+    userEvent.click(await screen.findByText(/show toast/i))
+    expect(await screen.findByText(/payment failed/i)).toBeInTheDocument()
+    userEvent.click(await screen.findByText(/discard/i))
+    expect(await screen.queryByText(/payment failed/i)).toBeNull()
```
diogoredin: No need for await here 💅

diogoredin (approved):
Awesome 🚀

PR #1370, opened 2022-02-17
Introduce tests and fix memory leaks of useTransition

### LINKED ISSUE #1225: Investigate if Dialog component is cancelling all async tasks on unmount
## Problem description
The Dialog component seems to have a memory leak. Most event handlers are removed when the component is unmounted but [useTransition](https://github.com/TicketSwap/solar/blob/main/src/hooks/useTransition.ts) doesn't seem to.

Screenshot 2021-10-21 at 17 41 08

## Steps to reproduce
1. Go to Sirius
2. Run any test suite where a Dialog is used multiple times (and hence mounted and unmounted multiple times);

## Expected behaviour
All async tasks should be cancelled upon unmounting, the component shouldn't leak.

## Additional context
I believe we don't see this memory leak in Solar tests because these are fairly fast (we don't do any data mocks, loading states etc.), which would otherwise give time for these async tasks to try to update the state of unmounted instances.

diogoredin:
# Description

@Roohn and I look into the memory leaks we had in our project tests and found most of them were originated in the `useTransition` hook of the Dialog.

If you have access you can see the warnings categorized [here](https://github.com/TicketSwap/sirius/issues/5839). The improvements this PR introduces reduces the memory leaks from 20 to 5 in the project.

**Important changes made:**
- The transitioning state is removed since we can use the "state" state to determine that;
- `rAF` inside a `rAF` is not necessary if we `useLayoutEffect`. Contrary to `useEffect`, which fires after the browser has painted, `useLayoutEffect` fires before the painting phase of the browser. One `rAF` call is still necessary to schedule the layout update for the current frame. This issue has however the issue of not being available on the server-side and producing an ugly warning. A hack from [here](https://github.com/reduxjs/react-redux/blob/d16262582b2eeb62c05313fca3eb59dc0b395955/src/components/connectAdvanced.js#L40) is applied to silence the warning, but we can keep the previous approach of nested `rAF` if you prefer (let me know in the comments);

https://reactjs.org/docs/hooks-reference.html#uselayouteffect
https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame
https://blog.jakuba.net/request-animation-frame-and-use-effect-vs-use-layout-effect/

- `useCallback` does not have any cleanup functions like `useEffect`, contrary to what we had in the code. Instead, we now clear the timers and scheduled `rAF` when the component unmounts (needs an empty dependency array to fire only once and on unmount).

## Changes

- [x] This has been done
- [ ] I still need to to this

## How to test

**To test the useTransition**
- Checkout this branch
- `$ yarn dev`
- Check the storybook for `ContentTransition`

**To check in your project**
- `$ yarn prepublishOnly`
- Copy the package inside `dist`
- Place it inside the project using solar (`node_modules`)
- Check the transitions

**To check the memory leaks are reduced:**
- Same as above
- Paste this [code](https://gist.github.com/diogoredin/35910541eba889b3da13604307ef187e) in your setup tests file
- Check the debug.log and see the memory leaks are less than stated [here](https://github.com/TicketSwap/sirius/issues/5839). 

## To be notified

@TicketSwap/frontend 

## Links

Fixes #1225


src/hooks/useTransition/index.ts
```diff
@@ -0,0 +1,115 @@
...
+  in: on,
+  defaultTransitionState = TransitionState.EXITED,
+  timeout,
+  onExited = () => {},
+  onEntering = () => {},
+  onEntered = () => {},
+  onExiting = () => {},
+}: useTransitionParameters): [TransitionState, boolean, boolean] => {
+  if (typeof on !== 'boolean') throwError('in')
+  if (typeof timeout !== 'number') throwError('timeout')
```
Roohn: Nit: you can remove these sinds we already have typescript type checking 🤔

Roohn (approved):
LGTM!!!

src/hooks/useTransition/index.ts
diogoredin: I considered that as well (removing this gets us to 100% coverage btw 😛 ) however I don't get typescript warnings for wrong args when using the component in Sirius, so I think this still has value.

diogoredin: Actually not, if this produces a runtime error as well it should be enough (because it's what we have now). Testing!

bobbonius (approved):
😎 Awesome Diogo!

src/hooks/useTransition/index.ts
diogoredin: Fixed ✅

diogoredin:
Waiting on https://github.com/TicketSwap/solar/pull/1378/files, `yarn prepublishOnly` stopped working and I want to test this one last time before merging.

PR #1379, opened 2022-02-23
Introduce Radio button

bobbonius:
# Description

Introducing `Radio`, `Label`, `RadioWithLabel` and `RadioGroup` component.

## How to test

- Checkout this branch
- `$ yarn storybook`
- Go to the Radio stories and check if everything is working as expected.

## Screenshots

image

image

image

## To be notified

@TicketSwap/frontend 

## Links

Solves a TO-DO item in sirius which you can find [here](https://github.com/TicketSwap/sirius/issues/5708)


src/components/Radio/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+  value,
+  name,
+  legend,
+}) => {
+  return (
+    <Fieldset>
+      {legend && <Legend>{legend}</Legend>}
+      {options.map(option => (
+        <RadioWithLabel
+          key={option.value}
```
RobVermeer: I think we can better user the `id` here =)

src/components/Radio/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  &:focus,
+  &:active {
+    outline: 0;
+    box-shadow: 0 0 0 ${space[4]} ${color.earthLightestAlpha};
+  }
+`
+
+export interface RadioProps {
```
diogoredin: The suggestion I was talking about in the frontend coffee:

```jsx
import { InputHTMLAttributes } from “react”;

type InputAttributes = Omit<
  InputHTMLAttributes<InputType>,
  “type” | “size” | “defaultValue” | “defaultChecked”
>;

export interface RadioProps extends InputAttributes {
  ref?: Ref<InputType>;
  checked: boolean;
  ...props
}
```

src/components/Label/index.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+    it('focusess on the accompanied input', () => {
+      const { getByLabelText, getByTestId } = render(
+        <>
+          <Label htmlFor="input">This is a label</Label>
+          <input id="input" data-testid="input" />
+        </>
+      )
+
+      const label = getByLabelText('This is a label')
+      const input = getByTestId('input')
```
diogoredin: **suggestion:** We can use the role instead of a test id. Doing this in tests helps us double-check our elements are accessible.

`screen.getByRole('textbox', {
  name: /this is a label/i
})`

src/components/Label/index.test.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+import { render } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { Label } from './'
+
+describe('Label', () => {
+  it('renders without crashing', () => {
+    const { container } = render(<Label htmlFor="input">This is a label</Label>)
```
diogoredin: **suggestion:** It's now recommended to use `screen` instead of extracting from the render method 💅

src/components/Label/index.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+    expect(container.firstChild).toBeInTheDocument()
```
diogoredin: **suggestion:** What do you think about instead checking for the label? With the current approach, we could add an extra div, make the label not be displayed and this test would still pass.

src/components/Label/index.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import styled from '@emotion/styled'
+import { device, fontSize, color } from '../..'
+
+export interface LabelProps {
```
diogoredin: **suggestion:** Maybe we should also extend the default input html type here, what do you think?

src/components/Radio/index.stories.tsx
```diff
@@ -0,0 +1,84 @@
...
+  return (
+    <Radio
+      id="radio1"
+      name="radio1"
+      onChange={event => {
+        console.log(event)
+        setChecked(event.target.value)
+      }}
+      value="radio1"
+      checked={checked === 'radio1'}
```
diogoredin: **nit-pick:** What I suggested the other day: data, booleans, functions 💅 

```jsx
id="radio1"
name="radio1"
value="radio1"
label="Radio 1"
checked={checked === 'radio1'}
onChange={event => {
  console.log(event)
  setChecked(event.target.value)
}}
```

diogoredin: Not going to pinpoint the other places, you get the idea 🙂

src/components/Radio/index.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+            ref={ref}
+            onChange={handleChange}
+            value="test"
+            checked={true}
+            id="test"
+            name="test"
+          />
+        )
+
+        userEvent.click(ref.current)
```
diogoredin: **suggestion:** What do you think about clicking on the label instead? (getByRole ...) It would resemble the user experience more closely.

src/components/Radio/index.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+              value: valueMock,
+            }),
+          })
+        )
+      })
+    })
+
+    describe('and it is already checked', () => {
+      it('does not fire an on change event', () => {
+        const handleChange = jest.fn(event => event.target.value)
```
diogoredin, line 91: **suggestion:** `jest.fn` should suffice. I think this was copy-paste from the other test that checked the callback. With the new check, this shouldn't be necessary there as well.

src/components/Radio/index.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+    expect(
+      screen.getByRole('presentation', {
+        name: /checkmark/i,
+      })
+    ).toBeInTheDocument()
+  })
+
+  describe('on click', () => {
+    it('fires an on change event when not checked before', () => {
+      const handleChange = jest.fn(event => event.target.value)
```
diogoredin: **nit-pick:** jest.fn should suffice now.

src/components/Radio/index.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+          id="test"
+          name="test"
+          disabled
+          label="test"
+        />
+      )
+
+      userEvent.click(ref.current)
+
+      expect(handleChange).toBeCalledTimes(0)
```
diogoredin: **suggestion:** not.toHaveBeenCalled 💅

src/components/Radio/index.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+    })
+  })
+})
+
+describe('RadioGroup', () => {
+  it('renders without crashing', () => {
+    const { container } = render(
```
diogoredin: **suggestion:** Using `screen`.

src/components/Radio/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+  grid-area: 1 / 1;
+  color: ${color.nova};
+  pointer-events: none;
+`
+
+const StyledInput = styled.input`
+  grid-area: 1 / 1;
+  background-color: ${color.nova};
+  height: 24px;
+  width: 24px;
```
diogoredin: **question:** Should this use space instead? or is that only for margins/paddings?

src/components/Radio/index.tsx
```diff
@@ -0,0 +1,166 @@
...
+    border-radius: ${radius.md};
+  }
+`
+
+export interface RadioWithLabelProps extends RadioProps {
+  label: string
+  ref?: React.Ref<HTMLInputElement>
+}
+
+export const RadioWithLabel: React.FC<RadioWithLabelProps> = forwardRef(
```
diogoredin, line 99: **suggestion:** Having all variations in the same file is okay for small components, but might get complex for large ones. What do you think of having a separate file for each one of the variations and then export them all from the index?

https://trusz.github.io/posts/barrel-pattern-in-typescript/

diogoredin (commented):
Awesome work @bobbonius and @RobVermeer. Left you some suggestions, let me know what you think 🙂

src/components/Label/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+  color: ${color.space};
+
+  @media ${device.tablet} {
+    font-size: ${fontSize[18]};
+  }
+`
+
+export const Label: React.FC<LabelProps> = ({ children, htmlFor }) => (
+  <StyledLabel htmlFor={htmlFor}>{children}</StyledLabel>
+)
```
glenngijsberts: This component works exactly the same as the default label, except for the fact that it's a styled component wrapped in a React component without any extra functionality. Why not export the styled component instead as we do with the `Text` component?

src/components/Radio/index.tsx
glenngijsberts: We could use `space` for that as well, we do that all the time so I don't see any problems with that.

src/components/Label/index.test.tsx
bobbonius: I'm not sure what you mean here @diogoredin

src/components/Label/index.tsx
bobbonius: I've tried this but it only gets more red. So not sure what to do here @diogoredin 

image

image

src/components/Radio/index.test.tsx
bobbonius: Test fail if i change it to just `jest.fn` @diogoredin

src/components/Label/index.test.tsx
Roohn: I agree, this doesn't really test anything 🤔

src/components/Radio/index.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+      ]}
+      onChange={e => setValue(e.target.value)}
+      value={value}
+      legend="This is a RadioGroup"
+      name="who-is-best"
+    />
+  )
+}
+
+describe('Radio', () => {
```
Roohn, line 37: It seems that you are missing a test to check if the `checked` prop is changed from false to true 🤔

src/components/Radio/index.test.tsx
```diff
@@ -0,0 +1,283 @@
...
+          onChange={handleChange}
+          value={valueMock}
+          checked={false}
+          id="test"
+          name="test"
+          label="test"
+        />
+      )
+
+      userEvent.click(ref.current)
```
Roohn: I think it is beter to click the label, because you already tested this on the radio without label.

Roohn:
I'm still not a big fan of having all these components separate but also inside other components.
For example: `<Radio/>` and `<Label />` are perfectly fine by me, because we are sticking to the html tags, but now there is a component that combines the two (`<RadioWithLabel />`) and there is a component that combines the component that combines the two (`<RadioGroup />`). RadioGroup can be done by adding RadioWithLabel components with the same name, like in normal HTML 🤔 . 
Concluding my rant, why do we want everything instead of ONLY the base components (radio and label) or ONLY the RadioGroup, because you can't use a radio on it's own, nor can you use a radio without a label 😬

src/components/Label/index.tsx
RobVermeer: Instead we want to have these styles inside the global styles. So we don't need the Label component at all.

bobbonius:
Merged via another PR

PR #1385, opened 2022-02-28
feat(Text): add TinyText component

### LINKED ISSUE #1377: Add a TinyText component
## Feature description
In designs (on Desktop) we want to have the option for a smaller text size than the SmallText, for example in dialog subtitles:
Screenshot 2022-02-22 at 11 57 12

At the moment there are already quite some places where we use this text size, but it's not an official component yet.

The text size would be `0.875rem` (14px) with a default line height of `1.5` as with all other text styles. On mobile the text size shouldn't change (14px is as low as we want to go).

Emacdyz:
# Description

Creation of the new TinyText 

## Changes

- [x] Exported new TinyText component, following design guidelines 
- [ ] Added related story

## How to test

- Checkout this branch
- `$ yarn run storybook`
- Go to TinyText story and check that everything is ok

## To be notified

@Marruk 

## Links
fixes https://github.com/TicketSwap/solar/issues/1377

### DIFF
--- a/src/components/Text/index.tsx
+++ b/src/components/Text/index.tsx
@@ -37,3 +37,13 @@ export const SmallText = styled.span`
     font-size: ${fontSize[16]};
   }
 `
+
+export const TinyText = styled.span`
+  ${baseTextStyles};
+  font-size: ${fontSize[14]};
+  color: ${color.spaceLight};
+
+  @media ${device.tablet} {
+    font-size: ${fontSize[14]};
+  }
+`
--- a/src/components/Text/index.stories.tsx
+++ b/src/components/Text/index.stories.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Text, SmallText } from '.'
+import { Text, SmallText, TinyText } from '.'
 
 export default {
   title: 'Components/Typography/Text',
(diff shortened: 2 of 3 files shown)


bobbonius:
You need to export it here in order for it to be useable in the build:

https://github.com/TicketSwap/solar/blob/a8eefd51047a8bff0c69b0ae1e27319687196dde/src/components/index.ts#L11

bobbonius (approved):
😎 🔍

src/components/Text/index.tsx
```diff
@@ -37,3 +37,13 @@ export const SmallText = styled.span`
...
   }
 `
+
+export const TinyText = styled.span`
+  ${baseTextStyles};
+  font-size: ${fontSize[14]};
+  color: ${color.spaceLight};
+
+  @media ${device.tablet} {
+    font-size: ${fontSize[14]};
```
diogoredin, line 47: Is this necessary if it's the same value as the other?

diogoredin (approved):
Left you a question, but LGTM

src/components/Text/index.tsx
Emacdyz: Yes it overrides the fontsize[18] that is coming from `${baseTextStyles}`

diogoredin: But we have font-size: `${fontSize[14]};` right after `${baseTextStyles};`, no? 🤔

diogoredin: This was a overwrite from the parent, got it 😅

PR #1386, opened 2022-03-01
Use own useIsomorphicLayoutEffect hook

diogoredin:
# Description

With the newly introduced `useIsomorphicLayoutEffect` hook, we can use it instead. The PR also removes @reach/utils as it was only used for this.

## Changes

- [x] This has been done
- [ ] I still need to to this

## How to test

- Checkout this branch
- `$ yarn storybook`
- Check Dialog component and ContentTransition

## Links

#1370 (hook recently introduced here)


bobbonius (approved):
Nice 😎

PR #1389, opened 2022-03-03
ci(ci.yml): prevent CI from running in some situations

bobbonius:
# Description

When a PR is a draft the CI should not run. It should only run when it's ready for review.

## Changes

- Used the same settings for Sirius in Solar

## How to test

- CI should not run if you create a draft PR

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,10 +10,14 @@ on:
   pull_request:
     branches:
       - main
+    types: [opened, ready_for_review, synchronize]
 
 jobs:
   tests:
     runs-on: ubuntu-latest
+    if: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -10,10 +10,14 @@ on:
   pull_request:
     branches:
       - main
+    types: [opened, ready_for_review, synchronize]
 
 jobs:
   tests:
     runs-on: ubuntu-latest
+    if: |
+        !contains(github.event.head_commit.message, '[SKIP TESTS]') &&
```
glenngijsberts, line 19: I don't think we need this `skip tests` stuff, since tests are quite fast in Solar :)

glenngijsberts (approved):
LGTM 👍

diogoredin (approved):

PR #1390, opened 2022-03-04
Correctly check mounted state

Roohn:
# Description
This PR will check if the dialog is still mounted before it calls the hide() function. It will fix the error you might see in tests if you unmount a dialog before calling `hide()`.
Next to that it will fix the `useIsMounted` hook and adds tests so that it doesn't break again.


diogoredin (approved):
Nice, LGTM!

PR #1397, opened 2022-03-09
Cancel entering animation on unmount

### LINKED ISSUE #1376: Check if transition is in progress before triggering new one
## Problem description
If you set a new view before the last one is finished transitioning we end up with multiple active views at the same time.

## Steps to reproduce
1. Pick the storybook example
2. Add one more view with a button that goes to another view
3. Check if you click on one button in one view and then on the other in the next view two views become active at the same time.

## Expected behaviour
We should wait for the previous view to become active before the next one or we should find a way to cancel the previous transition and only activate the new one.

## Additional conte

https://user-images.githubusercontent.com/1096800/154916279-21487b1f-70b7-4b22-88bd-6262cd866665.mp4

Roohn:
# Description
This PR fixes https://github.com/TicketSwap/solar/issues/1376
The bug was caused by not calling the handleExiting when a transition was in progress. When a component was still entering the view and then unmounted it wouldn't call the exit functions.

## Changes
- [x] Cancel the entering timeout/animation when a component is unmounted.

## How to test
1. Go to storybook
2. In the content transition story follow these views: `signup` > `signup with email` > `signup button` .
3. You should see the success view
4. Now click the back chevron twice in rapid succession.
5. You don't see the thing you see in screen recording AND there is no weird flickering on the dialog (that's why the transitioning check was there in the first place)

## Screenshots

OLD: 

https://user-images.githubusercontent.com/22071649/157454456-0f79064d-0ae4-45c3-8fca-827690deea53.mov

NEW: 

https://user-images.githubusercontent.com/22071649/157454474-ce5afcbf-f135-4d8f-945b-dce2a58232a7.mov

### DIFF
--- a/src/hooks/useTransition/index.ts
+++ b/src/hooks/useTransition/index.ts
@@ -71,7 +71,11 @@ export const useTransition = ({
   }, [onExiting, handleExited, timeout])
 
   useIsomorphicLayoutEffect(() => {
-    if (transitioning) return
+    if (state === TransitionState.EXITING) return
+
+    if (!on && typeof enteringTimerID.current !== 'undefined') {
+      clearTimeout(enteringTimerID.current)
+    }
 
     if (on && !prevOn) {
       setMounted(true)
@@ -84,7 +88,7 @@ export const useTransition = ({
     if (prevOn && !on) {
       exitingTimerID.current = handleExiting()
     }
-  }, [on, prevOn, handleEntering, transitioning, handleExiting])
+  }, [on, prevOn, handleEntering, transitioning, handleExiting, state])
 
   useEffect(() => {
     return () => {
--- a/src/components/ContentTransition/index.stories.tsx
+++ b/src/components/ContentTransition/index.stories.tsx
@@ -97,7 +97,13 @@ const Signup = (props: {
   </DialogBody>
 )
 
-const SignupEmail = ({}: { title: string; back: string }) => (
+const SignupEmail = (props: {
+  title: string
+  back: string
+  showSuccessView: (
+    event: React.MouseEvent<HTMLButtonElement, MouseEvent>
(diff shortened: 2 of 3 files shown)


src/hooks/useTransition/index.ts
```diff
@@ -71,7 +71,11 @@ export const useTransition = ({
   }, [onExiting, handleExited, timeout])
 
   useIsomorphicLayoutEffect(() => {
-    if (transitioning) return
+    if (state === TransitionState.EXITING) return
```
diogoredin, line 74: **suggestion:** We should have the state in the dependency array, otherwise we might be using an old value😅 Hope it still works with it synced, as it does with transitioning that also depends on this variable.

src/hooks/useTransition/index.ts
```diff
@@ -71,7 +71,11 @@ export const useTransition = ({
...
+
+    if (!on && typeof enteringTimerID.current !== 'undefined') {
```
diogoredin, line 76: **question:** When we click very quickly on the back we have two transitions potentially entering at the same time. That's why here we check if we have any scheduled entering animation already and cancel it to give place to the new one.

My doubt is why this isn't already done by the return below? They should clean up all animations on unmount, right? 💭

diogoredin (approved):
Awesome 💯 Thank you for looking into this 😁  Still, wrapping my head around it though, can you hopefully clarify my doubt?

src/hooks/useTransition/index.ts
Roohn: Yes but it won't reach the return below because it is still in the process of transitioning. So after the entering animation is done, it won't fire the exiting animation anymore because it thinks it already handled it by going into this useEffect (when it was returned early) 😅

PR #1398, opened 2022-03-10
feat(basebutton): add small baseButton size

### LINKED ISSUE #1368: Add smaller BaseButton variant to components
For designs we would love to have a smaller version of the `BaseButton` available to us, for links that need to be a bit more subtle.

Examples of where we could use them:
Screenshot 2022-02-16 at 14 14 52
(currently something custom)

Screenshot 2022-02-16 at 14 15 21
(currently a full-sized BaseButton which means that it gets too much attention compared to the "Total")

The button would be the same as the `BaseButton`, but with a text size that is the same as `SmallText` for both 
desktop and mobile.

Roohn:
# Description
Adds a small variant of the base button. Props are the same as we have on the button component, using `small` and `medium` sizes.

image

Fixes https://github.com/TicketSwap/solar/issues/1368

### DIFF
--- a/src/components/BaseButton/index.tsx
+++ b/src/components/BaseButton/index.tsx
@@ -1,7 +1,8 @@
 import React, { ReactNode } from 'react'
 import styled from '@emotion/styled'
-import { space, color, fontWeight, radius, device } from '../../theme'
+import { space, color, fontWeight, radius, device, fontSize } from '../../theme'
 import { baseTextStyles } from '../Text'
+import { css } from '@emotion/react'
 
 export enum ButtonVariant {
   info = 'info',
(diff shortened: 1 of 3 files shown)


src/components/BaseButton/index.tsx
```diff
@@ -1,18 +1,25 @@
...
+import { css } from '@emotion/react'
 
 export enum ButtonVariant {
   success = 'success',
   warning = 'warning',
   danger = 'danger',
 }
 
+export enum BaseButtonSize {
+  medium = 'medium',
```
glenngijsberts, line 15: Was thinking if we should call this `default` instead 🤔

glenngijsberts (approved):
LGTM

src/components/BaseButton/index.tsx
Roohn: Hmm yes that is also a good option. I was trying to keep it inline with the normal button though, maybe there are other people with opinions about it? 😄

bobbonius: I'm all for keeping it inline with the normal button

PR #1412, opened 2022-03-22
fix(colors): Change Pill colors to lighter 

Emacdyz:
# Description
Pill disabled check 

## How to test

- Checkout this branch
- Check in storybook
- Go to semantic color PR in sirius 
- Change the color of the pill on your dev tools to veryfy in dark mode as well

### DIFF
--- a/src/components/Pill/index.tsx
+++ b/src/components/Pill/index.tsx
@@ -41,7 +41,7 @@ export const Container = styled.span<PillProps>`
       : p.variant === 'warning'
       ? color.warning
       : p.variant === 'disabled'
-      ? color.inactive
+      ? color.inactiveBackground
       : color.action};
 `
 


src/components/Pill/index.tsx
```diff
@@ -41,7 +41,7 @@ export const Container = styled.span<PillProps>`
       : p.variant === 'warning'
       ? color.warning
       : p.variant === 'disabled'
-      ? color.inactive
+      ? color.strokeStrong
```
glenngijsberts: Hmm I think this should be `inactiveBackground` instead 🤔  and then `onInactive` would probably look better already?

Emacdyz: The production color of the pill was stardust 400 so `strokeStrong` was the one corresponding the most but i also hesitated between both because in term of naming the `inactiveBackground` is better so will change

src/components/Pill/index.tsx
```diff
@@ -29,7 +29,7 @@ export const Container = styled.span<PillProps>`
       : p.variant === 'warning'
       ? color.onWarning
       : p.variant === 'disabled'
-      ? color.onInactive
+      ? color.lightForeground
```
glenngijsberts: This should be `onInactive` now :) Looks already better!

glenngijsberts (approved):
If you change the color everything is good 👍

PR #1413, opened 2022-03-23
fix: Export old shadow value for Dark mode

diogoredin:
Since we don't have a correct color for this shadow yet.


glenngijsberts (approved):
Looking better 👍

glenngijsberts:
Maybe nice to create an issue for this to look into this problem together with Mark.

Roohn (approved):

PR #1424, opened 2022-03-31
fix: use correct colors for tooltip

glenngijsberts:
# Description

The colors of the tooltip weren't using the correct colors, so I've added proper background and foreground colors. Also scaled-down the tooltip, because it was a little too big imo.

@Marruk could you give feedback?

## How to test

- Checkout this branch
- `$ yarn storybook`
- Search for Tooltip
- Verify it's working

## Screenshots

Before
Screenshot 2022-03-31 at 14 58 33

After

Screenshot 2022-03-31 at 18 22 06
Screenshot 2022-03-31 at 18 22 23

### DIFF
--- a/src/components/Tooltip/index.stories.tsx
+++ b/src/components/Tooltip/index.stories.tsx
@@ -1,21 +1,28 @@
+import styled from '@emotion/styled'
 import React from 'react'
 import { Tooltip } from '.'
+import { space } from '../../theme'
+import { Button } from '../Button'
 
 export default {
   title: 'Components/Data Display/Tooltip',
 }
 
+const Grid = styled.div`
+  display: flex;
+  gap: ${space[16]};
+`
+
 export const Basic = () => (
-  <>
+  <Grid>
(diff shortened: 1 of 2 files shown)


Marruk:
I would be in favour of using `invertedBackground` and `invertedForeground`, the goal of the tooltip is to have a lot of contrast with whatever background it is on. @julimpaiva what do you think?

Julimpaiva:
> 

Makes sense, so would be the opposite of the bg that is in.

glenngijsberts:
@Julimpaiva @Marruk thanks for the feedback :) Applied it!

What do you both think of the size of the tooltip?

Roohn (approved):

Julimpaiva:
> 

Size looks good. It takes the size of text plus the padding right?

glenngijsberts:
> > 
> 
> Size looks good. It takes the size of text plus the padding right?

@Julimpaiva yes! thanks :)

PR #1428, opened 2022-04-01
fix(spinner): fix the spinner so it handles variants and correct colors

bobbonius:
BREAKING CHANGE: Added a new prop to `<Spinner />`

# Description

Added 

## Changes

- [x] This has been done
- [ ] I still need to to this

## How to test

- Checkout this branch
- `$ yarn dev`

## Screenshots

Capture anything that helps others better understand the purpose of the pull request, for example UI updates or API output.

## To be notified

## Links

### DIFF
--- a/src/components/Spinner/index.tsx
+++ b/src/components/Spinner/index.tsx
@@ -6,10 +6,11 @@ import { color } from '../../theme'
 
 export interface SpinnerProps {
   size?: number
+  color?: string
 }
 
 const StyledSpinnerIcon = styled(SpinnerIcon)`
-  color: ${color.foreground};
+  color: ${props => props.color || color.foreground};
 `
 
 const rotate = keyframes`
@@ -29,8 +30,12 @@ const Rotate = styled.div`
   align-items: center;
 `
 
-export const Spinner: React.FC<SpinnerProps> = ({ size = 32, ...props }) => (
+export const Spinner: React.FC<SpinnerProps> = ({
+  size = 32,
+  color,
+  ...props
+}) => (
   <Rotate {...props} role="status">
-    <StyledSpinnerIcon size={size} />
+    <StyledSpinnerIcon size={size} color={color} />
   </Rotate>
 )
--- a/src/components/Button/index.tsx
+++ b/src/components/Button/index.tsx
@@ -47,6 +47,19 @@ interface StyledButtonProps extends ButtonProps {
   isSquare: boolean
 }
 
+const spinnerColor = {
+  primary: color.onAction,
+  secondary: color.action,
+  caution: color.failure,
+  success: color.onAction,
+  warning: color.onAction,
+  danger: color.onAction,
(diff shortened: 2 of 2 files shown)


src/components/Spinner/index.tsx
```diff
@@ -3,13 +3,36 @@ import styled from '@emotion/styled'
 import { keyframes } from '@emotion/react'
 import { Spinner as SpinnerIcon } from '../../icons'
 import { color } from '../../theme'
+import { ButtonVariant } from '../Button'
```
glenngijsberts: Probably nice if you create a `SpinnerVariant` instead.

src/components/Spinner/index.tsx
```diff
@@ -3,13 +3,36 @@ import styled from '@emotion/styled'
...
 
 export interface SpinnerProps {
   size?: number
+  variant?: string
```
glenngijsberts: This can be a `SpinnerVariant` type if you create a `SpinnerVariant`.

src/components/Spinner/index.tsx
```diff
@@ -3,13 +3,36 @@ import styled from '@emotion/styled'
...
+      ? color.onFailure
+      : props.variant === ButtonVariant.facebook
+      ? color.onAction
+      : props.variant === ButtonVariant.inverted
+      ? color.info
+      : props.variant === ButtonVariant.apple
+      ? color.invertedForeground
+      : props.variant === ButtonVariant.google
+      ? color.darkBackground
+      : color.onAction};
```
glenngijsberts: Hmmm, doubt if we should use variants for the spinner. For example, these variants are focused on buttons but we also use a spinner on different pages as a "page loader". What we also can do is keep the `foreground` color by default, and then inside the button component specify a color for each variant. Other than the buttons I can't really name a place where we would use those variants 🤔

glenngijsberts (commented):
Left some feedback 👍  Maybe nice to have other opinions as well @TicketSwap/frontend

src/components/Spinner/index.tsx
```diff
@@ -6,10 +6,11 @@ import { color } from '../../theme'
 
 export interface SpinnerProps {
   size?: number
+  color?: string
 }
 
 const StyledSpinnerIcon = styled(SpinnerIcon)`
-  color: ${color.foreground};
+  color: ${props => props.color || color.invertedBackground};
```
glenngijsberts: A color shouldn't be a background right? Why not default to `color.foreground`?

bobbonius: @glenngijsberts the spinner then looks weird in an input component.. I'm wasn't quite sure to what it should default but foreground did not seem to be it.

bobbonius: Yes, you're right!

glenngijsberts (approved):
Nice work 👍

PR #1484, opened 2022-06-02
Make it possible to focus OneTimeCodeInput on mount

glenngijsberts:
# Description

Currently, it's not possible to focus the first input of the OneTimeCodeInput on mounting. This PR will make that possible.

## How to test

- Checkout
- Open Storybook
- Go to the `WithFocusOnMount` story of `OneTimeCodeInput` component
- Verify it will auto focus the first input on mounting

### DIFF
--- a/src/components/OneTimeCodeInput/index.stories.tsx
+++ b/src/components/OneTimeCodeInput/index.stories.tsx
@@ -69,6 +69,37 @@ export const ShortCode = () => {
   )
 }
 
+export const WithFocusOnMount = () => {
+  return (
+    <Dialog on>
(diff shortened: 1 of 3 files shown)


glenngijsberts:
It's better to use the native `autoFocus` attribute on your inputs, instead of using refs to focus something. But in general, autoFocusing is not a good practice for accessibility according to https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/287854abd066704e2a9964da597e7ab7f6f7e2ad/docs/rules/no-autofocus.md. Will do some more research to do this before making a change like this.

PR #1503, opened 2022-06-20
Add z-index to Toast component

glenngijsberts:
# Description

Currently, the toast component is showing behind the dialog overlay. We want to make sure that a toast is always visible, so by adding a z-index we can make sure it overlaps the dialog overlay.

## How to test

- Checkout this branch
- Open Storybook
- Go to the With Dialog Toast story of the toast component
- Verify the toast is being shown above the dialog overlay

## Screenshots

Screenshot 2022-06-20 at 12 07 48

## To be notified

@Marruk


bobbonius (approved):
Great use of the 
image

index

PR #1505, opened 2022-06-23
Make it possible to use string values for Typescript enums

glenngijsberts:
# Description

This PR makes it possible to use string values for Typescript enums. Currently, if you would use Solar you need to use `ButtonSize.large` when you want to declare the size of the `Button` component. This PR makes it possible to use the value of the enum as well, so like this: `<Button size="large">Hello world</Button>`.

The enums of the components are working with a lowercase value, and I think that makes sense for enums that are being used for props. The enum `TransitionState` is more for config of the transitions (and `useTransition`) and I think it makes more sense to use uppercase values for that. This means that we have a breaking change in that.

Let me know what you think!

## Changes

- [x] Make it possible to use string values for enums
- [x] Removed unneeded type in Logo/Path
- [x] Removed unneeded default transition in OnboardingDialog

## How to test

- Checkout this branch
- `$ yarn dev`
- Verify the related components (Buttons, Alerts, Dialogs, etc.) are working as expected in Storybook

### DIFF
--- a/src/components/Alert/index.tsx
+++ b/src/components/Alert/index.tsx
@@ -33,7 +33,7 @@ export enum AlertVariant {
 export interface AlertProps {
   title?: string
   action?: AlertAction
-  variant?: AlertVariant
+  variant?: keyof typeof AlertVariant
   children: ReactNode
 }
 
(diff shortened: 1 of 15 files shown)


bobbonius (approved):
image

Emacdyz (approved):
Nicee! 🛸

PR #1506, opened 2022-06-24
fix(Alert): fixed horizontal padding in Alert being on the wrong side

### LINKED ISSUE #1483: Alert contents have their horizontal padding switched
See screenshot:
Screenshot 2022-06-02 at 11 13 55

The content div within an alert has wrong `padding-inline` values, applying it to the left instead of the right.

Marruk:
# Description
The horizontal padding for the content in Alerts was switched around, having spacing on the left and not on the right as it should.

## Links
Fixes https://github.com/TicketSwap/solar/issues/1483


bobbonius (approved):
Thank you @Marruk

glenngijsberts (approved):

PR #1508, opened 2022-06-24
Fix vertical alignment in the Card component

glenngijsberts:
# Description

This PR will make sure to respect the vertical alignment prop of the Card component, see first commit for the bugfix and see second commit for an example in Storybook.

## How to test

- Go to https://ticketswap.github.io/solar/?path=/story/components-surfaces-card--with-text-and-subtitle
- Verify the adornments are not vertical aligned to the top (although they have that prop set)

- Checkout this branch
- `$ yarn storybook`
- Go to the `With Text and Subtitle` story of the Card component
- Verify the adornments are vertical aligned to the right position

## Screenshots

Screenshot 2022-06-24 at 20 18 39


bobbonius (approved):
Nice Glenn! 😎

PR #1536, opened 2022-07-14
Textarea component

RobVermeer:
# Description
This PR introduces the textarea component, so we don't need to use the same input component for textareas.

## Changes

- [x] add Textarea component

## How to test

- Checkout this branch
- `$ yarn storybook`
- See Components/Inputs/Textarea


bobbonius (approved):
Sexy component 😎

PR #1538, opened 2022-07-15
fix(textarea): fix textarea resize option

RobVermeer:
# Description

We want to make sure that you can't resize the textarea horizontally.


bobbonius (approved):
resize me 😎

PR #1547, opened 2022-07-22
feat(callAll): Export the callAll util as we also want to use it in sirius

viktorrenkema:
# Description
In Sirius, we also have a callAll util. To keep things leaner, I want to remove it in sirius and get it from solar instead. After this PR is approved, I'll import it in sirius and replace the old util in sirius.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,5 @@
 export * from './theme'
 export * from './hooks'
-export { normalizeChars } from './utils'
+export { normalizeChars, callAll } from './utils'
 export { globalStyles } from './global-styles'
 export * from './components'


bobbonius:
You can use `yarn commit` for semantic commits or change it manually to:

`'feat(callAll): Export the callAll util as we also want to use it in sirius'`

bobbonius:
@viktorrenkema You need to change the commit, not the title of the PR 😬

viktorrenkema:
> @viktorrenkema You need to change the commit, not the title of the PR 😬

Haha woops, thanks. Any reason why Solar requires semantic commits while Sirius does not?

bobbonius:
> > @viktorrenkema You need to change the commit, not the title of the PR 😬
> 
> Haha woops, thanks. Any reason why Solar requires semantic commits while Sirius does not?

@viktorrenkema Solar is open source, so for good practice reasons I guess

bobbonius (approved):
😎

glenngijsberts:
> > > @viktorrenkema You need to change the commit, not the title of the PR 😬
> > 
> > 
> > Haha woops, thanks. Any reason why Solar requires semantic commits while Sirius does not?
> 
> @viktorrenkema Solar is open source, so for good practice reasons I guess

and because it's a (NPM) package, which needs versioning. Semantic commits makes it easy to determine versions, generate release notes, etc.

glenngijsberts (approved):
👍 

Maybe nice to make your commit title a bit smaller, as you can see here it's already a bit too big for GitHub itself, and some other tools for git are also making it harder to read the commit message. For example, you could say `feat(callAll): export the callAll util` and then add the Sirius part in the body.

Also, since Solar is something on it's own (and open source) it can be a bit weird to mention Sirius here, because for people outside the company Sirius is unkown, so they are missing context.

Screenshot 2022-07-25 at 11 47 06

viktorrenkema:
> feat(callAll): export the callAll util

Nice one, didn't think of that. Changed :)

PR #1557, opened 2022-08-01
SelectV2

bobbonius:
# Description

Introducing the new `<Select />` component. 

## How to test

- Checkout this branch
- `$ yarn && yarn storybook`
- Test the behavior of a native select on desktop and mobile (for reference: [mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select))

## Screenshots

image


viktorrenkema:
@bobbonius getting this when testing on mobile iOS 15.6.

When I select any of the blank options, the option value does display in the Select value so the values are there, just not showing.

There also seems to be a background peeking behind the border-radius, this isn't the case for the v1 version.

image

RobVermeer:
> @bobbonius getting this when testing on mobile iOS 15.6.
> 
> When I select any of the blank options, the option value does display in the Select value so the values are there, just not showing.
> 
> There also seems to be a background peeking behind the border-radius, this isn't the case for the v1 version.
> 
> image

Fixed both issues!

viktorrenkema (approved):
LGTM visually!

glenngijsberts:
Think it would be nice to add some tests to this, in order to cover all cases (+ check for Typescript).

src/components/SelectV2/Select.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const [currentValue, setCurrentValue] = useState(
+    props.value || props.defaultValue || options[0].value
+  )
+  const currentOption = options.find(
+    option => option.value === currentValue
+  ) as OptionProps
+
+  return (
+    <Container hasStartAdornment={Boolean(startAdornment)}>
+      {!props.disabled && (
```
glenngijsberts, line 149: We can destruct this directly from props right?

src/components/SelectV2/Select.tsx
```diff
@@ -0,0 +1,201 @@
...
+  startAdornment = null,
+  rounded = false,
+  variant = 'default',
+  onChange,
+  'aria-label': ariaLabel,
+  ...props
+}: SelectProps) => {
+  const selectboxRef = useRef<HTMLSelectElement>(null)
+  const [currentValue, setCurrentValue] = useState(
+    props.value || props.defaultValue || options[0].value
```
glenngijsberts, line 141: We can destruct `defaultValue` as well right? Since we're using other `value` variables in this file I can see why we don't want to destruct that, although that's also possible.

src/components/SelectV2/DesktopDropdown.tsx
```diff
@@ -0,0 +1,161 @@
...
+export const DesktopDropdown = ({
+  'aria-label': ariaLabel,
+  options,
+  value,
+  onChange,
+  variant,
+}: DesktopDropdownProps) => {
+  const listboxRef = useRef<HTMLDivElement>(null)
+  const [open, setOpen] = useState(false)
+  const [focussed, setFocussed] = useState(
```
glenngijsberts, line 37: This indicates to me that it's a boolean, although it's an index right? Maybe `focussedElement` would be a better name, or anything other that indicates it's an index.

bobbonius: We also have a `selected` prop and I think it makes more sense to have both `focussed` and `selected`. Talked to Glenn and decided to keep it as it is.

src/components/SelectV2/DesktopDropdown.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+const Wrapper = styled.div`
+  position: absolute;
+  inset: 0;
+  display: none;
+  user-select: none;
+
+  @media (hover: hover) {
+    display: block;
+  }
```
glenngijsberts, line 15: 😍

glenngijsberts (approved):
Really nice work Brob 😎 🥳

PR #1558, opened 2022-08-03
One time code - Fixes

RobVermeer:
# Description
There were 2 issues with the one time code.

1. You couldn't make sure that the numeric keyboard on mobile was present when filling in the numbers (if you want to)
2. When doing a backspace when there was nothing in the field it would not move the cursor


bobbonius (approved):
Works as expected now! 😎

PR #1564, opened 2022-08-12
fix: Use action color for `a` and `button`

glenngijsberts:
# Description

See files changed. The `a` and `button` are representing actions, so they should use an action color, instead of the theme independent brand colors (such as earth).


bobbonius (approved):
😎

PR #1574, opened 2022-08-15
fix:checking if selectedItem exists on Dropdown select input

Emacdyz:
# Description
The bug is actually an edge case. The dateRange picker is opening the selects (year, months) but in the case of the `months` select, the value array is empty when dateRange is on different years. The util to get the month list was filtering the month between `startDate` and the `endDate`. So anything that will be superior to the `startDateMonth` (for ex. 6) but inferior to the `endDateMonth` (for ex. 1), which would return an empty array and therefore cause the error.  
 
https://ticketswap.atlassian.net/browse/TISSUES-799

### DIFF
--- a/src/utils/dates/index.ts
+++ b/src/utils/dates/index.ts
@@ -43,13 +43,31 @@ const getMonths = (locale: string, dateRange?: { start: Date; end: Date }) => {
   }
 
   if (dateRange) {
+    const startYear = dateRange.start.getFullYear()
+    const endYear = dateRange.end.getFullYear()
+
+    // check if the date range is on different years
+    if (endYear !== startYear) {
+      const monthsOfTheYearAfterStartMonth = months.filter(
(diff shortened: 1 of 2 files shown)


diogoredin (approved):

src/utils/dates/index.ts
```diff
@@ -43,13 +43,16 @@ const getMonths = (locale: string, dateRange?: { start: Date; end: Date }) => {
...
-      month =>
-        Number(month.value) >= dateRange.start.getMonth() &&
-        Number(month.value) <= dateRange.end.getMonth()
+    const monthsOfTheYearAfterStartMonth = months.filter(
+      month => Number(month.value) >= dateRange.start.getMonth()
+    )
+
+    const monthsOfTheYearBeforeEndMonth = months.filter(
+      month => Number(month.value) <= dateRange.end.getMonth()
     )
```
diogoredin: Nice, can we also add a test? 👀

PR #1576, opened 2022-08-18
Introduce `onBrand` and some other style adjustments

glenngijsberts:
# Description

Together with Mark, I went over some things, and we decided to introduce an `onBrand` color so that it gets easier to have themes. Made a fix for the Card component, by changing the text to lightForegroundMuted if it's on an image (good reminder that texts on images should always use the independent `light` version). Also removed the background-image from the secondary variant button. Design is working on new buttons, where the background-image gets removed anyway, but at the moment it doesn't benefit the secondary button at all, and is only causing style issues, so we decided to remove it for the time being.

## How to test

- Checkout this branch
- `yarn storybook`
- Verify everything is working as expected

### DIFF
--- a/docs/Guidelines/Colors/data.ts
+++ b/docs/Guidelines/Colors/data.ts
@@ -73,6 +73,12 @@ export const semanticColors = [
         hex: '#00B6F0',
         foreground: color.lightForeground,
       },
+      {
+        label: 'On Brand',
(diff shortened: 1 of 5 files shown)


RobVermeer (approved):
🖌️

viktorrenkema (approved):

PR #1584, opened 2022-08-23
Introduce Dropzone Component

diogoredin:
# Description
This PR introduces a stateless Dropzone component using the [HTML5 Drag and Drop specification](https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API).

## Usage
```jsx
<Dropzone
  title="Upload PDF or Apple Wallet tickets"
  subtitle="If your original file contains multiple tickets make sure to upload all of them."
  action="Drop files here or click here to select"
  dropTitle="Release"
  accept={['.pdf', '.txt']}
  onFileChange={file => console.log(file)} // can set state here
  onUnacceptedFileChange={() => {}} // can do something in case its wrong
/>
```

## Changes

- [x] This has been done
- [ ] I still need to to this

## How to test

- Checkout this branch
- `$ yarn storybook`
- Look for `Dropzone`
- Test in different browsers / devices

## Screenshots
**Desktop**
Screenshot 2022-08-23 at 17 17 01

**Mobile**
Screenshot 2022-09-05 at 14 24 13

## To be notified
@Julimpaiva @mon-d @TicketSwap/frontend 

## Links
[HG-2167](https://ticketswap.atlassian.net/browse/HG-2167)


viktorrenkema:
I noticed some weird behaviour for the **restricted** variant:
- it doesn't seem to accept any pdf i **drop**, yet clicking the button and then uploading does work as it doesn't run onUnacceptedFileChange (as expected)
- however, I would expect a console log when I click the button in above variant, yet as visible in recording below, nothing shows.


https://user-images.githubusercontent.com/49148610/186352920-0bcb68b1-d76d-43bc-8f62-30d1a9802f2b.mov

Marruk:
Nice! I guess we should add this to Figma as well right @Julimpaiva?

Also, the screenshots don't look like the font is Proxima Nova, but maybe that's because it's in the Storybook environment.

diogoredin:
@viktorrenkema There was an issue in the condition not considering the extension for certain file names, it's fixed now 🙏 

Screenshot 2022-08-26 at 07 55 34

@Marruk Yes, indeed, should be good once in use there I think.

src/components/Dropzone/index.tsx
```diff
@@ -0,0 +1,94 @@
+import React, { ChangeEvent, useRef } from 'react'
```
RobVermeer: Let's create a `Dropzone.tsx` with all this and then inside `index.ts` export this component and all types that need to be exported (see `InputV2` as example)

src/components/Dropzone/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { SmallText } from '../Text'
+import { BaseButton } from '../BaseButton'
+import ILLUSTRATION from './files.svg'
+
+interface DropzoneProps {
+  title: string
+  subtitle: string
+  action: string
+  ariaLabel?: string
+  accept?: Array<string>
```
RobVermeer: `string[]` is what we use on most places

src/components/Dropzone/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+  onFileChange: Function
```
RobVermeer: I think it is nice to type arguments and return value as well (if any). So more like this: `onFileChange: (event: SyntaticEvent) => void` (just an example, no idea wht this needs) (in anyway the arrow function definition is nicer that way imo, more explicit)

src/components/Dropzone/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const onSelectFile = (e: ChangeEvent<HTMLInputElement>) => {
+    if (e?.target?.files?.length) {
+      onFileChange(e.target.files[0])
+    }
+  }
+
+  return (
+    <DropArea data-testid="droparea">
+      <>
+        <Illustration src={ILLUSTRATION} />
```
RobVermeer: Images need width and height property on it, to calculate size and aspect ration before css is loaded (always the actual size of the image). Also an alt is needed, `alt=""` for images without meaning.

src/components/Dropzone/index.tsx
diogoredin: oops, fixed!

src/components/Dropzone/index.tsx
diogoredin: Applied 🙏

src/components/Dropzone/index.tsx
diogoredin: Indeed, I added it as a placeholder and forgot to improve it 🙈 fixed it!

src/components/Dropzone/index.tsx
diogoredin: Interesting! What happens if we don't do the calculation with CSS? Will it display the image in fullsize until the Javascript is loaded in the client side? 👀

diogoredin: Fixed btw 🙏

src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,214 @@
...
+      <DropOverlay
+        data-testid="dropoverlay"
+        isDragging={isDragging}
+        onDragOver={onDragOver}
+        onDragLeave={onDragLeave}
+        onDragExit={onDragExit}
+        onDrop={onDrop}
+      >
+        <img src={ILLUSTRATION} alt={dropTitle} />
+        <h2>{dropTitle}</h2>
```
viktorrenkema, line 197: Same as the h4 comment

src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,214 @@
...
+    background-color: ${color.actionBackground};
+    border: 2px dashed ${color.earth};
+  }
+`
+
+const Illustration = styled.img`
+  cursor: pointer;
+`
+
+const Title = styled(H4)`
```
viktorrenkema, line 88: This is now always an h4. Would it be an idea to accept a prop for the `as` attribute so the user of the Dropzone component can configure how this heading should be interpreted? Maybe an h4 doesn't make sense in all situations.

src/components/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,214 @@
...
+  space,
+  radius,
+  transition,
+  fontWeight,
+  device,
+} from '../../theme'
+import { H4 } from '../Heading'
+import { SmallText } from '../Text'
+import { BaseButton } from '../BaseButton'
+import ILLUSTRATION from './files.svg'
```
viktorrenkema: I dont think the svg is centered 🤔 
image

viktorrenkema:
I don't fully get the drop behaviour, sometimes i am able to drag the file anywhere in the overlay area (especially when I go 'around the sides' like in the video), but often it will also cancel the drop when I drag it to the same area from a different starting point.

I'd expect the entire overlay area to not cause a dismissal of the drop action.

https://user-images.githubusercontent.com/49148610/188565014-801e2076-3386-497e-a4d3-318680ebee9e.mov

Actually when typing this out, I figured it's the illustration and text that are not part of the drop area and therefore cancel the event. Working as expected when I remove those elements:

https://user-images.githubusercontent.com/49148610/188566632-a49149a6-4966-4934-96c7-20a6c122e7e2.mov

Maybe we can do something with z-indexes or other workarounds to have the drop-area be still 'on top' and not canceling the event?

src/components/Dropzone/Dropzone.tsx
diogoredin: Looking at other components that aren't specifically made for holding text (e.g. `Card`) I think it's okay to make assumptions about the hierarchy on the page. But let me know if you think otherwise @TicketSwap/frontend I'm also okay with adding the prop.

src/components/Dropzone/Dropzone.tsx
diogoredin: Looking at other components that aren't specifically made for holding text (e.g. `Card`) I think it's okay to make assumptions about the hierarchy on the page. But let me know if you think otherwise @TicketSwap/frontend I'm also okay with adding the prop.

diogoredin:
@viktorrenkema The issue with the drop behaviour was solved, can you take another look? 🙏

src/components/Dropzone/Dropzone.tsx
diogoredin: Nice catch! Fixed 🙏

viktorrenkema (approved):
🫡

PR #1585, opened 2022-08-24
feat(CardWithPeriod): Added new period section 

Emacdyz:
# Description
After the ongoing event epic, we figured we could do some improvements in how we show the dates. This PR add the option to show the date period within the event cards

## Changes

- [x] Added new `CalendarAlt` icon
- [x] Added new date section in case there is a period passed as a prop
- [x] Added corresponding story with all the different options 

## How to test

- Start storybook 
- Look for card > "With period" story 
- Have a look of the real implementation in Sirius 

## To be notified
@Marruk 

## Links
https://ticketswap.atlassian.net/browse/APO-2145


src/components/Card/index.tsx
```diff
@@ -203,11 +206,23 @@ const TopLeftAdornment = styled.div`
...
+`
+
+const PeriodText = styled(TinyText)`
+  padding-inline-start: ${space[4]};
+
+  :first-letter {
+    text-transform: capitalize;
+  }
+`
+
```
glenngijsberts: You shouldn't use padding to create space between elements, but margin 😬 

Btw if you use flexbox, you could use `gap` instead 😄 

```
const Period = styled.div`
  display: flex;
  align-items: center;
  gap: ${space[4};
`

File: src/components/Card/index.tsx [hunk 2]
```diff
@@ -342,6 +358,13 @@ const Card: React.FC<CardPropTypes> = ({
           {leftAdornment && <LeftAdornment>{leftAdornment}</LeftAdornment>}
 
           <TextContent hasImage={hasImage}>
+            {period && (
+              <Period>
+                <CalendarAlt size={16} color={color.mars} />
+                <PeriodText color={color.mars}>{period}</PeriodText>
+              </Period>
+            )}
```
glenngijsberts: I wonder if we should have a specific `period` prop for a generic component like a `card`. For a lot of use cases, it doesn't make sense to have a `period` prop. I think it would be better to have a `header` prop (consistent with the `footer` one) and use that to display a period label (in our own project, for example, creating an `EventCard`).

src/components/Card/index.tsx
```diff
@@ -203,11 +206,23 @@ const TopLeftAdornment = styled.div`
   }
 `
 
+const Period = styled.div`
+  display: inline-flex;
+  align-items: center;
+`
+
+const PeriodText = styled(TinyText)`
```
Marruk: Could you make this have a semi bold font weight please?

src/components/Card/index.tsx
```diff
@@ -203,11 +206,23 @@ const TopLeftAdornment = styled.div`
...
+  padding-inline-start: ${space[4]};
+
+  :first-letter {
+    text-transform: capitalize;
```
Marruk: What is the reason you are doing this?

Marruk:
Yay looks great!

I don't think you touched it, but the subtitle in the image variant has `foregroundMuted` as the color, which makes it quite dark in light mode. Could you make that `lightForegroundMuted`?

As we discussed the Card component could use a big clean-up, will try to make a proposal soon.

src/components/Text/index.tsx
```diff
@@ -41,7 +41,7 @@ export const SmallText = styled.span`
 export const TinyText = styled.span`
   ${baseTextStyles};
   font-size: ${fontSize[14]};
-  color: ${color.foregroundSubtle};
+  color: ${props => props.color ?? color.foregroundSubtle};
```
glenngijsberts: Why do we need this `color` prop? You can also use the styled component to overwrite this, right? This is not really consistent with the other text elements like H*, Text, etc.

src/components/Card/index.tsx
```diff
@@ -342,6 +358,13 @@ const Card: React.FC<CardPropTypes> = ({
           {leftAdornment && <LeftAdornment>{leftAdornment}</LeftAdornment>}
 
           <TextContent hasImage={hasImage}>
+            {period && (
+              <Period>
+                <CalendarAlt size={16} color={color.mars} />
+                <PeriodText color={color.mars}>{period}</PeriodText>
```
glenngijsberts: Maybe a question for Mark: Is the text supposed to be red? I think the iOS app is not red text.

Marruk: Yeah, on the iOS app we have non-bold titles and in that case the red drowns out the title. On Web that's less of an issue and the red gives it some more graphical swung. It can go either way, but the difference is intended.

src/components/Card/index.tsx
Emacdyz: Yes, indeed it is better, I updated it to something more general.

src/components/Card/index.tsx
Emacdyz: I now do it from the util itself, so not necessary anymore. The auto generated date like today or tmw were not capitalize and i thought first it would be easier to do it from here but at the end not. I removed it

src/components/Card/index.tsx
Emacdyz: I now have the possibility to custom the color of the header, it would default to Tiny Text color if not precised

src/components/Text/index.tsx
Emacdyz: I think for a header that could be custom to any purpose with an icon and a color, it would make sense to have it like this so you can have it green, red, yellow or whatever color suits you best.

src/components/Card/index.tsx
```diff
@@ -342,6 +360,15 @@ const Card: React.FC<CardPropTypes> = ({
...
 
           <TextContent hasImage={hasImage}>
+            {header && (
+              <Header>
+                {header.icon}
+                {header.text && (
+                  <HeaderText color={header.color}>{header.text}</HeaderText>
+                )}
+              </Header>
+            )}
```
glenngijsberts: This is still inconsistent with the footer, the right adornment, etc. 

I think we should have the same implementation as the footer prop, and then you can pass JSX to that prop in order to render for example the icon and the text.

Emacdyz: I thought this would be a bit repetitive in Sirius to do as there will be quite some places where we need to add that period section with icon and text but indeed, this is very general so everybody can use it as they like. Updated!

glenngijsberts: Imagine you want to use this for events, and you want to give all event cards a period header, I think it's a good idea to introduce an `EventCard` component in your own codebase, that's basically implementing this `Card` component with a few set props, such as the period :) 

If we have something like that, it would make it very easy to update the event cards across our whole platform, instead of replacing all the cards one by one.

src/components/Card/index.tsx
```diff
@@ -203,11 +204,16 @@ const TopLeftAdornment = styled.div`
   }
 `
 
+const Header = styled.div`
+  display: flex;
+  align-items: center;
+  gap: ${space[4]};
```
glenngijsberts: I don't think we need this styling, I think this should become a normal `header` on line `352` without a styled component. You should apply this styling to the element that you are passing to the header prop 😄

src/components/Card/index.stories.tsx
```diff
@@ -23,6 +24,12 @@ const Grid = styled.div`
   }
 `
 
+const Header = styled.div`
```
glenngijsberts: Ideally, this should be a `header` element, and the `leftAdornment` inside the Card should be a regular div. Not sure if it makes sense that the left adornment is a header 😄

glenngijsberts (approved):
Nice job 👍

PR #1607, opened 2022-09-15
fix: Made gradient backdrop for cards with images stronger to make sure text remains readable

Marruk:
With the new Browse Events cards the text became unreadable with some images:
image

As a quick fix I made the gradient stronger, planning to make bigger changes (including the hover state) when looking at the Card altogether.

It now looks like this (disregard the different text label, that is unchanged):
Screenshot 2022-09-15 at 12 11 10


bobbonius (approved):
😎

Issue #1608, opened 2022-09-15
Dark mode is not working for components using portals


glenngijsberts:
## Improvement description
Components, such as Toasts and Dialogs, are not working in dark mode. This has to do with the portals that are being used, that are rendering the components outside of the root of the playground and so they don't have a parent with the `data-theme="dark"` property.

## Expected behaviour
Portals should be either rendered inside the root, or we should find another way to make those components work in dark mode.

Issue #1507, opened 2022-06-24
Inverted colors for Toasts


Marruk:
## Feature description
Not too sure why https://github.com/TicketSwap/solar/pull/1470 was done to revert https://github.com/TicketSwap/solar/pull/1452 (inverted colors) for Toasts. With those changes there isn't a lot of contrast between the background colors anymore, making them hard to miss.

While playing around with this with in a local Storybook I noticed that the available stories were applying colors wrong (no difference between light and hard mode, an override of `foreground` for the text). Maybe that's where it went wrong?

The goal of this is to have dark Toasts in light mode and light Toasts in dark mode, using the `invertedBackground` and `invertedForeground`.

## Additional context
Paging @RobVermeer @bobbonius to maybe provide context on why they reverted the inverted colors so we can have a discussion here if needed.

PR #1609, opened 2022-09-15
Display `Toast` with inverted colors

glenngijsberts:
# Description
Design would like to use inverted colors for our toasts, so that they are better visible. This was already done before, and reverted, but this was probably reverted because the colors weren't correctly displayed in storybook (see https://github.com/TicketSwap/solar/issues/1507). This is because the toasts are displayed in a portal, and storybook is not handling them really great. Added #1608 to improve this.

## How to test

- Checkout this branch
- `yarn storybook`
- Go to the `Toast > Persistent` story
- Click Show Toast
- Verify the toast is displayed with a dark background color
- Enable dark mode in storybook
- Click right mouse on the Toast and inspect HTML
- Add `data-theme="dark"` to the `html` tag
- Verify the toast is having a light background

## Screenshots
Screenshot 2022-09-15 at 13 47 45

Screenshot 2022-09-15 at 13 48 09

## To be notified

@Marruk

### DIFF
--- a/src/components/Toast/index.stories.tsx
+++ b/src/components/Toast/index.stories.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import { Toast, ToastProvider, useToast } from './'
 import { Button, ButtonSize, ButtonVariant } from '../Button'
+import { BaseButton } from '../BaseButton'
 import { Select } from '../Select'
 import styled from '@emotion/styled'
 import { CheckmarkRounded } from '../../icons'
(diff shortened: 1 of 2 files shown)


Marruk:
Nice, thanks!

Buttons look a bit weird, will be fixed with the button changes that we have in mind. If super easy we could go for a text button for now.

bobbonius (approved):
😎

glenngijsberts:
> Nice, thanks!
> 
> Buttons look a bit weird, will be fixed with the button changes that we have in mind. If super easy we could go for a text button for now.

@Marruk Done 👍 It's btw not something inside the Toast itself, so when we use a Toast we need to make sure to properly style the content inside it.

PR #1616, opened 2022-09-20
Revamp `Button` component

glenngijsberts:
# Description

The design team would like to introduce a new type of button, that should replace the secondary & caution button variant. This new button introduces a `weight` prop (with `full` and `outlined` as values) that makes it possible to render all variants with only an outline instead of a background color (see `With Outlined Weight` story in Storybook).

Besides the addition of this new prop, the button got a small makeover. A few variants got replaced (with the new `weight` prop, or other alternatives), a few variants got removed (such as the social media buttons and the warning button) and the `background-image` has been removed from all variants. One of the goals of this revamp was also to make the styles a bit more maintainable. 

The `loading` prop has been removed, which can now be achieved using the `startAdornment` (which was `leftAdornment` before) prop. This is done to make it more consistent with the new Inputs, where this is also done. The square functionality has been made explicit, to have more control in when a button should be displayed as a square.

In order to make it possible to gradually migrate to this new version, I've created a V2 variant of the button. Check the stories in Storybook to verify how it works.

## How to test

- Checkout
- `yarn`
- `yarn storybook`
- Go to `ButtonV2`

## Screenshots

Screenshot 2022-09-20 at 15 26 43

## To be notified

@TicketSwap/design


Marruk:
Yay nice!

Couple of remarks/discussion points/random other things:
- Are we using the "Warning" variant somewhere in our Product at the moment?
- The Facebook / Google / Apple variants feel a bit weird to include in the Design System as we're only using them in the log in modal (I think). 
- Regardless of that the Google and Apple variants look the same with the full weight, don't think that's intended
- In the "With states" page the one with the loading indicator is misaligned, that makes me a bit scared
- I'm a bit confused by the "With Square" variants, shouldn't it become a square automatically with only the icon showing? It feels more logical to me to base that on the elements being present or not instead of a separate property. Or is the square stuff not an actual property?
- It would be nice if the loading indicator wouldn't take up extra space so the buttons remain the same size while toggling between loading / not loading states.
- For the outlined variant it would be nice to have the "background" variant of the color become the background on a hover, but not worth it if it's a hassle.

Don't think any of these are very pressing, can pick this all up later as we can put a bit more time in. This is a bit of a braindump too, feel free to ignore stuff.

glenngijsberts:
@Marruk yeah I was already thinking about ditching the external variants, so I'll actually do that (and replace it with a variant created inside our own codebase). Will also remove the warning one, because we don't use it anywhere, and I assume that you want to get rid of it if we don't use it?

About the misalignment of the button, that's because the old buttons were having a `vertical-align` property, but IMO the placement/alignment should not be a responsibility of the element itself but the parent. Adding a display flex to the parent can already fix that, so I did that in storybook to make it look a bit better.

About the other points, will have a look at that!

glenngijsberts:
> I'm a bit confused by the "With Square" variants, shouldn't it become a square automatically with only the icon showing? It feels more logical to me to base that on the elements being present or not instead of separate property. Or is the square stuff not an actual property?

I gave this some thought and personally, I think it's best to have the square prop explicit. This way, you can control yourself if you want to have the button displayed as a square or normal. It makes the component API a bit more clear IMO instead of having that functionality implicit and using a prop (the adornment) that normally serves as an adornment but then acts as a child.

With the old button, it's also not possible to have a button with only an icon/spinner without being a square, so this also makes that possible. A downside is that you still can pass children + square, and if the child is not an icon it breaks the design. In a next PR I want to add some types to prevent this.

> It would be nice if the loading indicator wouldn't take up extra space so the buttons remain the same size while toggling between loading / not loading states.

Will pick this up in a next PR!

> For the outlined variant it would be nice to have the "background" variant of the color become the background on a hover, but not worth it if it's a hassle.

Done!

Marruk:
> This way, you can control yourself if you want to have the button displayed as a square or normal.

I think my point is mostly that I don't really want people to control themselves whether they want a square or normal button. Maybe I'm missing use cases, but to me the look of the buttons should be uniform across the product and they should either all be square or all be rounded. Providing an option opens the door to inconsistencies based on "I think this looks better".

glenngijsberts:
@Marruk 

> I think my point is mostly that I don't really want people to control themselves whether they want a square or normal button.

But this is already the case with the current button component. If you pass only an icon as children to the current button, it will not make the button square (and in this context, it means that the width & height is equal, not related to the border-radius). If you only pass a `leftAdornment` without children it does make it a square, which is weird and is hidden logic IMO. This PR only makes it more explicit. This way the component API is more clear, which can also be valuable for people that are using this outside of TicketSwap. When we add better types in the next PR, we can force consistency while keeping a clear API.

See the screenshots below for the different use cases btw. If you don't want to have those differences, or not have both kinds of variants, we should come up with a replacement for either one of them.

Screenshot 2022-09-20 at 23 48 55

Screenshot 2022-09-20 at 23 55 12

bobbonius:
I'm not the biggest fan the `outlined` button types on dark background such as in your example. I thought it was hard to see and also hard to determine it's a button. I do like the refactor a lot. Nice work. I would like to see the button inside sirius and see how it behaves there also bearing a11y in mind.

Marruk:
> If you pass only an icon as children to the current button, it will not make the button square (and in this context, it means that the width & height is equal, not related to the border-radius). If you only pass a leftAdornment without children it does make it a square, which is weird and is hidden logic IMO.

Ah yeah I wasn't aware of that hidden logic, definitely agree that an explicit property is better than that. I would still argue that it's even better to have consistent behaviour: if a button only shows an icon (in whatever way) it's a square. The loading indicator in the second screenshot is made with that "it replaces the content and keeps the button the same size" change. Happy to merge for now though (since that loading indicator stuff isn't implemented anyway).

bobbonius (approved):
Nice job Glenn

image

PR #1631, opened 2022-10-04
Vikrob/collapsible

viktorrenkema:
@RobVermeer

### DIFF
--- a/src/components/CollapsibleV2/Collapsible.tsx
+++ b/src/components/CollapsibleV2/Collapsible.tsx
@@ -1,0 +1,193 @@
+import React, { ReactNode, useState, useRef, useId, useEffect } from 'react'
+import styled from '@emotion/styled'
+import {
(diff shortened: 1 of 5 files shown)


src/components/CollapsibleV2/Collapsible.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+const Details = styled.div<DetailsProps>`
+  position: relative;
+  text-align: start;
+  width: 100%;
+  background-color: ${color.elevatedBackground};
+  border-radius: ${radius.md};
+`
+
+const Button = styled.button<ButtonProps>`
```
RobVermeer, line 34: Nit: Types need to be above the thing that uses the type

viktorrenkema: Check ✔️

PR #1645, opened 2022-10-21
Introduce secondary action in DropZone

diogoredin:
# Description

This PR allows the Dropzone to have a secondary action. e.g. a help button that can be used to present useful information.

## Changes

- [x] This has been done
- [ ] I still need to to this

## How to test

- Checkout this branch
- `$ yarn dev`

## Screenshots

Screenshot 2022-10-21 at 11 41 41
Screenshot 2022-10-21 at 11 41 51
Screenshot 2022-10-21 at 11 43 19
Screenshot 2022-10-21 at 11 43 29

## To be notified
@TicketSwap/frontend 

## Links
[JIRA HG-2168](https://ticketswap.atlassian.net/browse/HG-2168)

### DIFF
--- a/src/components/Dropzone/Dropzone.tsx
+++ b/src/components/Dropzone/Dropzone.tsx
@@ -1,4 +1,10 @@
-import React, { useState, ChangeEvent, DragEvent, useRef } from 'react'
+import React, {
+  useState,
+  ChangeEvent,
+  DragEvent,
+  useRef,
+  ReactNode,
+} from 'react'
 import styled from '@emotion/styled'
 import { css } from '@emotion/react'
 import {
@@ -29,6 +35,7 @@ export interface DropzoneProps {
   subtitle: string
   action: string
   mobileAction?: string
+  secondaryAction?: ReactNode
   dropTitle: string
   accept?: string[]
   multiple?: boolean
@@ -43,6 +50,10 @@ interface InformationStyles {
 }
 
 interface DropAreaStyles {
+  variant: DropzoneVariant
+}
+
+interface DesktopUIStyles {
   variant: DropzoneVariant
 }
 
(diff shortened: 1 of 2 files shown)


viktorrenkema (commented):
Doesn't look like it's going well on mobile
image

viktorrenkema (commented):
Was this how it was design by Design? To me a couple things feel weird:
- the 'view example' is super close to the button, i'd easily misclick/tap the blue btn instead
- the fact that there's only ~10px between them makes it feel out of balance with the padding below the view example button.
image

diogoredin:
Thanks for the feedback! The goal here is to do something similar to what we have in production right now:
Screenshot 2022-10-21 at 13 15 59


Indeed I checked and the small variant (with everything horizontal) is not used on mobile, so we now have:

**Default Variant**
_Desktop_
Screenshot 2022-10-21 at 13 34 58

_Mobile_
Screenshot 2022-10-21 at 13 34 04


**Small Variant**
_Desktop_
Screenshot 2022-10-21 at 13 33 27

_Mobile_
Screenshot 2022-10-21 at 13 33 43

diogoredin:
cc @Julimpaiva

viktorrenkema (approved):
Got it, it already looks much better now!

A final nitpick is that in your 2nd image of your last comment, this now feels unbalanced. But I'll let @Julimpaiva be the judge of that :)

Screenshot 2022-10-21 at 13 39 28

Julimpaiva:
> Got it, it already looks much better now!
> 
> A final nitpick is that in your 2nd image of your last comment, this now feels unbalanced. But I'll let @Julimpaiva be the judge of that :)
> 
> Screenshot 2022-10-21 at 13 39 28

@diogoredin thats true! So the outer padding (from icon to border and "view example" to border should be 23px. And the inside paddings, from one element to other ("drop files" to "view example" should be 13px.

rest looks good!

PR #1655, opened 2022-11-02
Add missing title and back props to DialogBody

diogoredin:
# Description

Most of the times this is used in conjunction with `ContentTransition` which expects a title `string` and `back` key to be present (they are used in the `onClick` callback).

## Changes

- [x] This has been done
- [ ] I still need to to this

## How to test

- Checkout this branch
- `$ yarn dev`

## To be notified
@TicketSwap/frontend

### DIFF
--- a/src/components/Dialog/index.tsx
+++ b/src/components/Dialog/index.tsx
@@ -237,6 +237,8 @@ const Body = styled.div`
 
 type DialogBodyProps = {
   children: ReactNode
+  title?: string
+  back?: string
 }
 
 export const DialogBody: React.FC<DialogBodyProps> = ({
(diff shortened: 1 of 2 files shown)


viktorrenkema (approved):

src/components/Dialog/index.tsx
```diff
@@ -237,6 +237,8 @@ const Body = styled.div`
 
 type DialogBodyProps = {
   children: ReactNode
+  title?: string
+  back?: string
```
glenngijsberts, line 241: Can we add a storybook story for this?

bobbonius (approved):

src/components/Dialog/index.tsx
bobbonius: Done

PR #1656, opened 2022-11-03
Add check if option or currentOption exists

viktorrenkema:
# Description
Should prevent errors when no currentOptions are passed to the component.

## How to test
- [ ] Confirm the Select button still works as expected

### DIFF
--- a/src/components/SelectV2/Select.tsx
+++ b/src/components/SelectV2/Select.tsx
@@ -143,9 +143,8 @@ export const Select = ({
   const value =
     typeof externalValue !== 'undefined' ? externalValue : internalValue
 
-  const currentOption = options.find(
-    option => option.value === value
-  ) as OptionProps
+  const currentOption =
+    options.find(option => option.value === value) || ({} as OptionProps)
 
   return (
     <Container hasStartAdornment={Boolean(startAdornment)}>


Emacdyz (approved):
LGTM!

src/components/SelectV2/Select.tsx
```diff
@@ -191,7 +191,7 @@ export const Select = ({
       >
         {options.map(({ startAdornment, ...option }, index) => (
           <option key={index} {...option}>
-            {option.label || option.value}
+            {option?.label || option.value}
```
RobVermeer: If `option` would be undefined, it would also be true for `option.value`. That said, I don't think option can be undefined here, since it's a rest operator, it would always default to an empty object 🤔

src/components/SelectV2/Select.tsx
```diff
@@ -204,7 +204,7 @@ export const Select = ({
       >
         {variant !== 'default' && currentOption.startAdornment}{' '}
         {!variant.startsWith('short') &&
-          (currentOption.label || currentOption.value)}
+          (currentOption?.label || currentOption.value)}
```
RobVermeer: This one can be undefined, but has the same issue, might be better to default the currentOption to an empty object.

```ts
const currentOption = options.find(
    option => option.value === value
  ) || { } as OptionProps
```

Something like this.

viktorrenkema: Ah yes this approach makes more sense, thanks!

src/components/SelectV2/Select.tsx
```diff
@@ -204,7 +203,7 @@ export const Select = ({
...
```
RobVermeer:
```suggestion
          (currentOption.label || currentOption.value)}
```

You don't need the optional chaining here, since currentOption is always an object.

RobVermeer (approved):

PR #1664, opened 2022-11-14
File type checking dropzone

diogoredin:
# Description

If we have a file name with dots, `Screenshot 2022-11-14 at 11.10.58.png`, we should only consider the last one.

## Changes

- [x] This has been done
- [ ] I still need to to this

## How to test

- Checkout this branch
- `$ yarn dev`
- Upload a file with multiple dots
- See it is logged in your console log as an accepted file

## Screenshots
Screenshot 2022-11-14 at 17 19 39

## To be notified

@TicketSwap/frontend


RobVermeer (approved):
👍

PR #1665, opened 2022-11-15
chore(Panel): replace shadow with stroke

viktorrenkema:
# Description
Moving forward, elements with no events like Panels will have a stroke instead of a shadow. If a Panel should also include e.g. a click event, we recommend using a Card instead.

## Changes
- [ ] Removed the shadow CSS property
- [ ] Added a stroke

## How to test
- [ ] Checkout this branch
- [ ] Run storybook, confirm the shadow is replaced by a border.

### DIFF
--- a/src/components/Panel/index.tsx
+++ b/src/components/Panel/index.tsx
@@ -1,7 +1,6 @@
 import styled from '@emotion/styled'
 import {
   radius,
-  shadow,
   color,
   space,
   device,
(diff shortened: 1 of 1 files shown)


RobVermeer (approved):

src/components/Panel/index.tsx
```diff
@@ -16,7 +15,7 @@ export interface PanelContentProps {
 export const Panel = styled.div`
   ${fullBleed};
   background-color: ${color.background};
```
Marruk: This might not be needed anymore without a shadow?

glenngijsberts (approved):

src/components/Panel/index.tsx
viktorrenkema: Agreed, we do set the background but on the panel content. Checked, doesn't seem to do much now.

PR #1666, opened 2022-11-15
Revamped chip component 

Emacdyz:
# Description
In the Browse Events epic, we aim to revamp the filters' design. To do so, new components needed to be created. The next steps will be to implement them on the browse page in sirius. 

## Changes

- Created new Chip component 
- Extended the Chip component to be a select 
- Extender the Chip component to be a toggle button 
- Added related test 

## How to test

- Checkout this branch
- `$ yarn storybook`
- Search for Chip
- Check all the stories related 
- Make sure the test passes 

## Links
https://ticketswap.atlassian.net/browse/APO-2294

### DIFF
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -1,0 +1,129 @@
+import React, { ButtonHTMLAttributes, ReactNode } from 'react'
+import styled from '@emotion/styled'
+import { css } from '@emotion/react'
+import {
+  space,
+  color,
+  device,
+  lineHeight,
+  fontSize,
+  fontWeight,
+} from '../../theme'
+
+const Content = styled.span`
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  pointer-events: auto;
+`
+
+interface ContainerProps {
+  active: boolean
+  compact: boolean
+  hasStartAdornment: boolean
+}
+
+const Container = styled.span<ContainerProps>`
+  --_accentColor: ${color.inactive};
+  --_backgroundColor: ${color.background};
+
+  &::before {
+    content: '';
+    position: absolute;
+    inset: -2px;
+    border: 2px solid var(--_accentColor);
+    border-radius: ${space[32]};
+    pointer-events: none;
+  }
+
+  display: inline-flex;
+  align-items: center;
+  position: relative;
+  cursor: pointer;
+  padding-inline: ${space[16]};
+  padding-block: ${space[8]};
+  gap: ${space[4]};
+  background-color: var(--_backgroundColor);
+  border-radius: ${space[32]};
+  color: var(--_accentColor);
+  max-width: ${space[144]};
+  line-height: 20px;
+  font-size: ${fontSize[14]};
+  font-weight: ${fontWeight.semiBold};
+  user-select: none;
+
+  & > * {
+    pointer-events: none;
+  }
+
+  & ${Content} {
+    pointer-events: auto;
+  }
+
+  ${({ active }) =>
+    active &&
+    css`
+      --_accentColor: ${color.action};
+      --_backgroundColor: ${color.actionBackground};
+    `}
+
+  ${({ hasStartAdornment, compact }) =>
+    hasStartAdornment &&
+    compact &&
+    css`
+      > *:first-of-type {
+        display: none;
+      }
+
+      @media ${device.tablet} {
+        > *:first-of-type {
+          display: block;
+        }
+      }
+    `}
+
+  @media ${device.tablet} {
+    max-width: 240px;
+    gap: ${space[8]};
+    font-size: ${fontSize[16]};
+    line-height: ${lineHeight.copy};
+  }
+`
+
+export interface ChipProps {
+  as?: 'button' | 'label'
+  active?: boolean
+  startAdornment?: ReactNode
+  endAdornment?: ReactNode
+  compact?: boolean
+}
+
+interface ButtonChipProps
+  extends ChipProps,
+    ButtonHTMLAttributes<HTMLButtonElement> {}
+
+export const Chip = ({
+  as = 'button',
+  active = false,
+  startAdornment,
+  endAdornment,
+  compact = false,
+  children,
+  ...props
+}: ButtonChipProps) => {
+  return (
+    <Container
+      as={as}
+      active={active}
+      compact={compact}
+      hasStartAdornment={Boolean(startAdornment)}
+      {...props}
+    >
+      {startAdornment}
+
+      <Content>{children}</Content>
+
+      {endAdornment}
+    </Container>
+  )
+}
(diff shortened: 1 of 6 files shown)


Marruk:
Yay awesome, looks great!

I found two hopefully small things:
* in the mobile version the left adornment makes the chip higher. Initially I had the idea that the left adornment would be hidden on the mobile version (if there is a label).
* I tried it on mobile Safari and the border radius for the outline didn't seem to work:
Screenshot 2022-11-15 at 17 05 33

The Chip could also use a better hover state but that's my fault, will add that the list of improvements for Solar.

Emacdyz:
> Yay awesome, looks great!
> 
> I found two hopefully small things:
> 
> * in the mobile version the left adornment makes the chip higher. Initially I had the idea that the left adornment would be hidden on the mobile version (if there is a label).
> * I tried it on mobile Safari and the border radius for the outline didn't seem to work:
>   Screenshot 2022-11-15 at 17 05 33
> 
> The Chip could also use a better hover state but that's my fault, will add that the list of improvements for Solar.

Will check on Sarafi, good one! For the height of the chip, it is actually an option. If we pass the compact props that will be based on device size, i think the height is correct (you can have a look on a compact story).

Marruk:
> For the height of the chip, it is actually an option. If we pass the compact props that will be based on device size, i think the height is correct (you can have a look on a compact story).

~~Ah nice!~~

~~I'm guessing it is technically possible to have a row of non-compact chips on mobile though right? In that case they will have different heights based on whether they have a start adornment or not:~~

Didn't realise the size of the icon isn't part of the component, my bad

glenngijsberts:
I think you can squash a few commits here and there. Your first commit is trying to import something that's not committed yet, and I think it's perfectly fine to squash those commits. Your Storybook changes are committed as a feature but those are more doc changes I guess and your test should be a test commit instead of a feature commit :)

src/components/Chip/Chip.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+interface ContainerProps {
+  active: boolean
+  compact: boolean
+  hasStartAdornment: boolean
+}
+
+const Container = styled.span<ContainerProps>`
+  --_accentColor: ${color.inactive};
+  --_backgroundColor: ${color.background};
```
glenngijsberts, line 28: I think I get why you're doing this but personally, I don't like it. I would rather call this `ChipAccentColor` instead of `_`. If you don't know that we have those variables in our theme it can be really weird to see a `_` before.

src/components/Chip/ChipWithSelect.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { Chip, ChipProps } from './Chip'
+
+const Content = styled.span`
+  select {
+    position: absolute;
+    opacity: 0;
+    inset: 0;
+    cursor: pointer;
+  }
+`
```
glenngijsberts: Why not style the `select` as styled component?

src/components/Chip/ChipWithSelect.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+interface OptionProps extends OptionHTMLAttributes<HTMLOptionElement> {}
```
glenngijsberts: If we don't add any extra properties, why extend something instead of using the thing that you're using to extend?

src/components/Chip/ChipWithSelect.tsx
```diff
@@ -0,0 +1,73 @@
...
+        <select
+          value={value}
+          onChange={event => {
+            setInternalValue(event.target.value)
+            onChange?.(event)
+          }}
+          {...props}
+        >
+          {options.map((option, index) => (
+            <option key={index} {...option}>
```
glenngijsberts: In the PR for the new Combobox we set the key like [this](https://github.com/TicketSwap/solar/pull/1615/files#diff-6ded7d0ece10dbe7a02884e7500135dfa041c03b2ea5acd98b3e50bb36de48caR132-R136). Maybe we can do that here as well?

src/components/Chip/index.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+      { label: 'Today', value: 'today' },
+      { label: 'Tomorrow', value: 'tomorrow' },
+      { label: 'This weekend', value: 'this-weekend' },
+    ]}
+  />
+)
+
+export const WithToggleControlled = () => {
+  const [checked, setChecked] = useState(false)
+  return (
```
glenngijsberts: Whitespace in between would be nice.

src/components/Chip/Chip.tsx
RobVermeer: It's a convention to start local variables with `_`.

src/components/Chip/ChipWithSelect.tsx
RobVermeer: It doesn't forward the props, so it's not possible (on purpose).

src/components/Chip/Chip.tsx
glenngijsberts: @RobVermeer Link to convention?

src/components/Chip/ChipWithSelect.tsx
glenngijsberts: Can you explain this? Why doesn't it forward the props, and what's not working?

RobVermeer: To be able to create a styled component from another component you have to pass the props to the wrapper of the child component.

So for example:

```js
const Parent = styled(Child)`
  background: red;
`

const Child = () => (
  <div>Hello My Child</div>
)
// ^^^^ This one is not red
```

```js
const Parent = styled(Child)`
  background: green;
`

const Child = (props) => (
  <div {...props}>Hello My Child</div>
)
// ^^^^ This one is green
```

src/components/Chip/Chip.tsx
RobVermeer: https://github.com/argyleink/open-props/blob/main/src/extra/buttons.css
See how they do it in `open-props`, created by one of the best CSSers Adam Argyle 🤓

glenngijsberts: But because they do something, doesn't mean we have to? 😄 Do you really think `_` is more clear? Personally, I think it's a weird "convention", and not descriptive.

src/components/Chip/ChipWithSelect.tsx
glenngijsberts: Zoomed with Rob, Rob thought the select was our Select component, and in that case, we can't style it as styled component. Because we're using a native select, we can safely style it directly 😄

Emacdyz: Sweet! Done!

src/components/Chip/Chip.tsx
Emacdyz: They also seem to do both `var(--_accent, var(--link));` so then it's a bit confusing why there is a difference. I find the `_` less easy to the eye but i have no strong opinion about it. We just need to stick to our convention..

RobVermeer (approved):

PR #1668, opened 2022-11-22
SelectV2 - Fix React warning for default value

RobVermeer:
# Description
This fixes a warning when you add a default value to the select instead of a value.
With this change, internally value is always used, but the user can both use default value or value and it would work uncontrolled or controlled.


bobbonius (approved):
😎

PR #1669, opened 2022-11-23
Add ability to make a dialog scrollable with fixed content height

viktorrenkema:
# Description
This PR enables the DialogBody to consume a height prop that triggers vertical scrolling.

If a `height` is passed, overflow-y is set to `auto` making the Body scroll if your content (in `children`) exceeds the height you set.

If no `height` is passed to the DialogBody component, it works as usual where the Dialog takes up the full height of the content.

When you need elements to stay fixed (e.g. a search panel), just add it the component or HTML _before_ or _after_ your DialogBody, similar as how we already use DialogFooter. The added Story is an example of this.

## How to test
- Checkout this branch
- `$ yarn dev`
- Test all the variants provided in Storybook for the Dialog on mobile & desktop (the "With scrollable body" is new)
- In `src/components/Dialog/index.stories.tsx`, play around with the content in `DialogBody` starting on line 210, and what happens when you add/remove/change the `height` prop
- Confirm that all the other variants in storybook still work as expected

## To be notified
@RobVermeer @glenngijsberts  Here's a proof of concept, interested to hear your thoughts if we think this would be a good reusable solution. I feel that allowing the user of the component to explicitly specify a (fixed) height is an intuitive way to control the behaviour of the dialog's body that can apply to various different use-cases.

## Screenshots
A dialog with longer content, as it would display today:
image

Add a height to the DialogBody:
image
image

### DIFF
--- a/src/components/Dialog/index.stories.tsx
+++ b/src/components/Dialog/index.stories.tsx
@@ -12,18 +12,21 @@ import { ContentTransition } from '../ContentTransition'
 import { Select } from '../Select'
 import { Button, ButtonVariant } from '../Button'
 import { Input } from '../InputV2'
-import { ChevronLeftAlt, CloseAlt } from '../../icons'
+import { ChevronLeftAlt, CloseAlt, MagnifyingGlass } from '../../icons'
 import { Textarea } from '../Textarea'
 import { BaseButton } from '../BaseButton'
+import { color, device, space } from '../../theme'
+import styled from '@emotion/styled'
+import { Card } from '../Card'
 
 const items = [
-  { value: 'de', name: 'German' },
-  { value: 'it', name: 'Italian' },
-  { value: 'nl', name: 'Dutch' },
-  { value: 'en', name: 'English' },
-  { value: 'hu', name: 'Hungarian' },
-  { value: 'fr', name: 'French' },
-  { value: 'es', name: 'Spanish' },
+  { value: 'de', name: 'German', flag: '🇩🇪' },
+  { value: 'it', name: 'Italian', flag: '🇮🇹' },
+  { value: 'nl', name: 'Dutch', flag: '🇳🇱' },
+  { value: 'en', name: 'English', flag: '🇬🇧' },
+  { value: 'hu', name: 'Hungarian', flag: '🇭🇺' },
+  { value: 'fr', name: 'French', flag: '🇫🇷' },
+  { value: 'es', name: 'Spanish', flag: '🇪🇸' },
 ]
 
 const BasicDialog = ({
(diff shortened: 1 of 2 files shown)


src/components/Dialog/index.tsx
```diff
@@ -239,18 +244,17 @@ type DialogBodyProps = {
   children: ReactNode
   title?: string
   back?: string
+  height?: string
```
glenngijsberts, line 247: I don't really like introducing a height prop for this. It makes the logic a bit hidden. Also, it leaves room for a lot of inconsistency. If you pass a height that is larger than the actual content, you have a lot of spacing left without any content, which looks weird as well. I would probably work with fixed sizes.

Screenshot 2022-11-30 at 11 37 20

src/components/Dialog/index.tsx
```diff
@@ -210,7 +210,12 @@ const BodyWrapper = styled.div`
   }
 `
 
-const Body = styled.div`
+interface BodyStyles {
+  height?: string
```
glenngijsberts, line 214: It probably makes more sense to pass a number as height, e.g. `375` instead of `375px`

src/components/Dialog/index.stories.tsx
```diff
@@ -153,6 +156,72 @@ export const WithLongBody = () => (
...
+            )}
+          </DialogHeader>
+          <Search>
+            <Input
+              id="preferredCurrency"
+              startAdornment={<MagnifyingGlass size={24} />}
+              aria-label="A search field"
+              placeholder="Search for your currency"
+            />
+          </Search>
```
glenngijsberts, line 209: If we have a header, body and footer I don't think it makes sense to have content that is not in one of those components. I would rather add this to the header, and then you probably need to adjust the header component a bit.

src/components/Dialog/index.tsx
viktorrenkema: Adding fixed heights could be a good way to go, so e.g. allowing users to pass a S M or L prop or something?

One concern I have is that with scrollable content, we want to convey to users that they can scroll. To do that, the dialog needs to be cut off at precise points so you see that there is content 'down', e.g. like below (where Mark also points it out).

If we only have a couple of predetermined heights, we don't give ourself a lot of control to decide where to cut the dialog off as content will differ.

image

viktorrenkema: Also, if you do pass a height that is larger than the content then it'll indeed look funny, but I don't think we're planning to pass dynamic heights, right? And when else would you give it a larger height than the content.

src/components/Dialog/index.tsx
viktorrenkema: Hah yeah I dont have a strong preference, I thought it'd be preferred to go the way to stay closer to the actual CSS (so including 'px'). But I'm cool with both.

src/components/Dialog/index.stories.tsx
viktorrenkema: Good one, yeah I agree this feels like I've taken an existing structure and messed it up a bit. Gonna look into it :)

viktorrenkema:
Closing this, going to approach it in a different more robust way.

PR #1673, opened 2022-11-30
Introduce type prop on OneTimeCodeInput

glenngijsberts:
# Description

Currently the OneTimeCodeInput can be used for various things. You could use it for numeric codes, but also codes with letters. Sometimes you know upfront that a code only consist out of numbers or letters. To improve the experience a bit, this PR adds a `type` prop that makes it possible to only accept letters, only accept numbers or both which is the default.

## How to test

- Checkout this branch
- `yarn storybook`
- Go to the stories of the OneTimeCodeInput
- Verify you can only input letters in the letters only story
- Verify you can only input numbers in the numbers only story
- Verify the other stories are still working the same


glenngijsberts:
We can use a `pattern` attribute instead.

PR #1677, opened 2022-12-09
Added name props to options of ChipWithSelect

Emacdyz:
In theory we would have a label everywhere, but in action (in sirius) the label is actually a `name` props as it is name of the period or the city or the category. While it could be updated in the category and period filter in the frontend, the backend is returning the list of location with city name instead of city labels. This way we support both.


RobVermeer (changes requested):
I think this is not the way to go, this change is specific to a repo that is not Solar. I think in this case you should just change the code that uses this component, instead changing the component itself =)

PR #1680, opened 2022-12-09
fix(icon): don't use position absolute on icons

RobVermeer:
# Description
Position relative means that the icon can be on top of other things, which is not really handy and also not needed.


bobbonius (approved):
😎

PR #1687, opened 2022-12-19
Add condition for short label

viktorrenkema:
# Description
The phone number select had issues as it was using the country code to determine which item in the select to display. Since multiple countries use the area code 1 (Canada, USA, Puerto rico etc) this lead to issues with the displayed value of the phone input.

## How to test
- Test PhoneInput

## Screenshots

Capture anything that helps others better understand the purpose of the pull request, for example UI updates or API output.

## To be notified

## Links


RobVermeer (approved):
🎙️

bobbonius (approved):

PR #1688, opened 2022-12-21
fix(chipWithSelect): inactive default option

Emacdyz:
# Description

A design request is to have the first element of the select (default value) as inactive 

## How to test

- Go to the storybook > chip with select controlled 
- default state should be inactive grey > if selecting something else > should become active blue

### DIFF
--- a/src/components/Chip/ChipWithSelect.tsx
+++ b/src/components/Chip/ChipWithSelect.tsx
@@ -44,7 +44,7 @@ export const ChipWithSelect = ({
   return (
     <Chip
       compact={compact}
-      active={Boolean(value)}
+      active={Boolean(value) && !options[0].value}
       startAdornment={startAdornment}
       endAdornment={<ChevronDown size={16} />}
       as="label"


bobbonius (approved):
😎

RobVermeer:
> default state should be inactive grey > if selecting something else > should become active blue

This doesn't seem to work for me, it will always be grey 🤔

RobVermeer (changes requested):
I don't think this is needed 🤔 (and it doesn't work currently)

As an example try this in storybook (the current one, without your changes)
```js
export const WithSelectControlled = () => {
  const [value, setValue] = useState('')

  return (
    <ChipWithSelect
      value={value}
      onChange={event => setValue(event.target.value)}
      options={[
        { label: 'Today', value: '' },
        { label: 'Tomorrow', value: 'tomorrow' },
        { label: 'This weekend', value: 'this-weekend' },
      ]}
    />
  )
}
```

In this case the one without value is grey, and if then changed to one with value, it will be blue. This is something you can fix in Sirius itself =) Just make sure that the default option doesn't have any value and then make sure that when you query for example that the empty value one will be the right one, if that makes sense 😄

PR #1695, opened 2023-01-09
fix(datePicker): corrected first day of the year in calendar

Emacdyz:
# Description

Following [this tissue](https://ticketswap.atlassian.net/browse/TISSUE-1290), made a little investigation. The way the calendar is built didn't include the fact that for some countries a week starts on Sunday instead of Monday. Therefore 1 Jan 2023 would show on a Monday instead of the correct sunday. This PR is correcting that bug, by finding the index of the day within the weekdays array, and unshifting the correct amount of day to show a correct calendar. 

## How to test

- Open storybook > datePicker story > open calendar > you will see that jan 1 is on monday in the calendar 
- checkout this branch 
- reopen the calendar > make sure January 2023 starts on a sunday 
- check the following years

### DIFF
--- a/src/components/DatePicker/CalendarDialog.tsx
+++ b/src/components/DatePicker/CalendarDialog.tsx
@@ -99,11 +99,17 @@ const CalendarDialog = ({
 
   const [selectedMonth, setSelectedMonth] = useState(initialMonth)
   const [selectedYear, setSelectedYear] = useState(initialYear)
-
   const daysInMonth = new Date(selectedYear, selectedMonth + 1, 0).getDate()
-  const firstDayOfMonth = new Date(selectedYear, selectedMonth, 1).getDay()
-
   const weekdays = getWeekdays(locale)
+
+  const firstDayOfMonth = weekdays.findIndex(
+    day =>
+      day ===
+      new Date(selectedYear, selectedMonth, 1).toLocaleDateString(locale, {
+        weekday: 'short',
+      })
+  )
+
   const months = getMonths(locale, dateRange)
   const years = getYears({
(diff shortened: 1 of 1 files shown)


glenngijsberts (approved):
Since this is fixing a current issue and works now I'll approve but it would be nice to have Sunday as the first day (so starting on the left with Sunday) instead of the last day if it's the first day of the week. Could we maybe also add a testcase for this?

CleanShot 2023-01-09 at 15 36 52@2x

Emacdyz:
> Since this is fixing a current issue and works now I'll approve but it would be nice to have Sunday as the first day (so starting on the left with Sunday) instead of the last day if it's the first day of the week. Could we maybe also add a testcase for this?
> 
> CleanShot 2023-01-09 at 15 36 52@2x

But for example if the first day of the year is thursday it would mean that all the weeks in the calendar starts on thursday which is a bit strange..

viktorrenkema:
I think Glenn means that for those countries where the Sunday is the official first day of the week, that the calendar also represents that. 

So this would mean that the calendar would start either with SUN or MON (also not for any other days, cause i'm assuming no country starts their week with the thursday), cause while your fix works, it will now still display the Sunday at the end of the list.

viktorrenkema (approved):

Emacdyz:
It is unfortunately not working out of the box as the weekdays is always generated in the same order, even being locale aware... I have a potential solution but this method is not supported on Firefox, so I will merge this now.[ I opened an issue ](https://github.com/TicketSwap/solar/issues/1696)to take it after my holidays

PR #1698, opened 2023-01-25
fix: Allow Dropzone to be used in forms

diogoredin:
# Description

If the button type is not specified the form is submitted immediately. Besides this, we should allow the button to be full width within forms.

## Changes

- [x] This has been done
- [ ] I still need to to this

## How to test

- Checkout this branch
- `$ yarn dev`

## Screenshots

Screenshot 2023-01-25 at 13 15 50

## Links

[HG-2169](https://ticketswap.atlassian.net/browse/HG-2169)

[HG-2169]: https://ticketswap.atlassian.net/browse/HG-2169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/BaseButton/index.tsx
+++ b/src/components/BaseButton/index.tsx
@@ -16,7 +16,8 @@ export enum BaseButtonSize {
   small = 'small',
 }
 
-export interface ButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
+export interface ButtonProps
+  extends React.ButtonHTMLAttributes<HTMLButtonElement> {
   leftAdornment?: ReactNode
   variant?: keyof typeof ButtonVariant
   disabled?: boolean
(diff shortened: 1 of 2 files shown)


viktorrenkema (approved):
Don't think this button will ever function as a 'submit' button, which it couldn't be with the current fix, so LGTM!

bobbonius (approved):
😎

src/components/BaseButton/index.tsx
```diff
@@ -68,15 +69,15 @@ const StyledButton = styled.button<ButtonProps>`
...
-      ${p =>
-        p.variant === ButtonVariant.success
-          ? color.success
-          : p.variant === ButtonVariant.danger
-          ? color.failure
-          : p.variant === ButtonVariant.warning
-          ? color.warning
-          : color.info};
+    box-shadow: 0 0 0 ${space[4]};
+    ${p =>
```
glenngijsberts: Can we rename this to `props` or destruct `variant`? 😅

src/components/Dropzone/Dropzone.tsx
```diff
@@ -276,6 +279,7 @@ export const Dropzone = ({
           <MobileUI>
             {variant === DropzoneVariant.light ? (
               <Button
+                type="button"
```
glenngijsberts, line 282: We should probably have a look at this button in general, but wouldn't make more sense to still make it possible to pass a type to this? I don't think we're using it as a submit button if other external people use a submit button it's breaking for them, right?

glenngijsberts: So I think it would be nice to take that into account OR add a breaking change in the commit message.

diogoredin: This button is always used to open the file manager from within the `Dropzone`, so I don't think anyone would be using it to submit a form. Nonetheless I have added a breaking commit.

src/components/BaseButton/index.tsx
diogoredin: fixed 🙏

PR #1705, opened 2023-02-13
feat(chip): added squared prop

Emacdyz:
# Description

With the genre filter coming up, design requested a squared version of the ChipWithToggle

## How to test

- Checkout this branch
- `$ yarn storybook`
- go to actions > chip > With toggle squared corners and check it looks different (more squared) but still work as the more rounded one 

## Links
https://ticketswap.atlassian.net/browse/APO-2611

### DIFF
--- a/src/components/Chip/Chip.tsx
+++ b/src/components/Chip/Chip.tsx
@@ -32,7 +32,7 @@ const Container = styled.span<ContainerProps>`
     position: absolute;
     inset: -2px;
     border: 2px solid var(--_accentColor);
-    border-radius: ${space[32]};
+    border-radius: ${space[8]};
     pointer-events: none;
   }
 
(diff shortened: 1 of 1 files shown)


src/components/Chip/Chip.tsx
```diff
@@ -26,13 +26,14 @@ interface ContainerProps {
 const Container = styled.span<ContainerProps>`
   --_accentColor: ${color.inactive};
   --_backgroundColor: ${color.background};
+  --_borderRadius: ${space[32]};
```
RobVermeer:
```suggestion
  --_borderRadius: ${({ squared }) => squared ? space[8] : space[32]};
```

I think this is a bit more clean, then you don't need a custom styled component. You do have to pass the squared prop to the main chip component, but I think this also just makes sense, so all the chip's can be "squared".

RobVermeer:
Do you have an example of a design with this squared chip? 🤔 To me it sounds a bit weird to have a chip which looks more like a normal button (why do we have a chip then??) 🤔 

Btw if this is a one-of, then the change to make it a customizable css variable, would be enough as well. So then there is no need to add a prop, just style the component inside Sirius (might not what we want right now though).

Emacdyz:
@RobVermeer, so Mark just told me that they made some updates and want all the chips to be more rectangelish, so that simplifies this PR!

viktorrenkema (approved):
🔴

bobbonius (approved):

PR #1710, opened 2023-03-09
Use logical properties for width and height

bobbonius:
# Description

This PR introduces more logical properties for height, width following [this](https://www.npmjs.com/package/stylelint-use-logical) package. The package used in one of our private repositories now supports: width, height, min-width, min-height, max-width, max-height.


RobVermeer (approved):
🏴

PR #1713, opened 2023-03-27
feat(selectv2): added css variables

bobbonius:
# Description

Added css variables to a variety of elements in order to more easily control styling in custom components

### DIFF
--- a/src/components/SelectV2/Select.tsx
+++ b/src/components/SelectV2/Select.tsx
@@ -78,18 +78,22 @@ interface DisplayValueProps {
 
 const DisplayValue = styled.span<DisplayValueProps>`
   grid-area: 1 / 1 / 1 / 4;
-  font-size: ${fontSize[18]};
-  line-height: ${space[56]};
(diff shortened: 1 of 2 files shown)


src/components/SelectV2/Select.tsx
```diff
@@ -78,18 +78,22 @@ interface DisplayValueProps {
 
 const DisplayValue = styled.span<DisplayValueProps>`
   grid-area: 1 / 1 / 1 / 4;
-  font-size: ${fontSize[18]};
-  line-height: ${space[56]};
-  block-size: ${space[56]};
-  padding-inline: ${space[16]} ${space[48]};
-  color: ${color.foreground};
+  font-size: var(--displayValueFontSize, ${fontSize[18]});
```
RobVermeer: I would prefix this with Select, since `displayValue` can be anywhere

bobbonius: Good idea

RobVermeer (approved):
✅

PR #1719, opened 2023-04-24
Introduce a new version for the Cover component

glenngijsberts:
# Description

In our main app we worked on a new cover for our entity pages, like event, city, landing & artist pages. This PR implements a new version of our cover component. It includes a bit more structure than our old cover component and is made re-usable for the variants designed by Mark.

There are a few things that are opinionated: the background image is blurred, the title and subtitle are strings and are having a white color text by default. Some other things are less opinionated because it's hard to have strict typing. For example, the meta info section is ideally a list of links and/or texts, but then it's not possible to have a custom link tag, for example if you are using Next.js with the `Link` tag. I think if you want to have consistency in those rows, you should create components in your own codebase (which is my plan as well). The stories in Storybook are showing how it can be implemented.

CleanShot 2023-04-23 at 20 55 08@2x
CleanShot 2023-04-23 at 20 51 39@2x

## How to test

- Checkout
- `$ yarn storybook`
- Check out the stories of `CoverV2`

## To be notified
@Marruk


src/components/CoverV2/Cover.tsx
```diff
@@ -0,0 +1,206 @@
...
+interface CoverContainerProps {
+  fullHeight: boolean
+}
+
+const CoverContainer = styled.div<CoverContainerProps>`
+  position: relative;
+  inline-size: 100%;
+  block-size: auto;
+  overflow: hidden;
+  background-color: var(--_coverBackgroundColor, ${color.darkBackground});
```
bobbonius, line 23: Nice 🔥

bobbonius (approved):
Nice work Glenn 😎

PR #1722, opened 2023-04-26
fix(theme): Fixed odd shadow/overlay colors in theme

Marruk:
This should fix two things:
* A light glow instead of a shadow in light mode
* Overlays not having a clear outline on the overlay in dark mode


glenngijsberts (approved):

glenngijsberts:
Thank you Mark

PR #1723, opened 2023-04-26
Carousel / CurratedContentCard

Emacdyz:
# Description

Introduced new content card for desktop/mobile. This will be used within the future carousel of the revamped feed. 

Design: https://www.figma.com/file/sr7w5I0c5AZnEzmRdbTgvM/User-Prefs?node-id=3680-84784&t=HjmR2P52CMPUBEZ4-0

The Carousel is also new. It can be of 1 or 2 columns. Arrow navigation on desktop and horizontal scroll on mobile desktop.

## How to test
- [ ] Seach for `ImageCard` in your storybook
- [ ] Introduced new ImageCard
- [ ] Desktop/tablet is an horizontal card 
- [ ] Mobile is a vertical card 
- [ ] Scaling hover effect on the pic of the dekstop card 
- [ ] Seach for `Carousel` in your storybook
- [ ]  Introduced new Carousel
- [ ] Desktop navigate with arrow and can have 2 columns 
- [ ] Mobile navigate with horizontal card
- [ ] Mobile is a vertical card 

## Links
https://ticketswap.atlassian.net/browse/APO-2981


src/components/CurratedContentCard/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  imageUrl: string
+  title: string
+  text?: string
+}
+
+export const CurratedContentCard = ({ imageUrl, title, text }: Props) => {
+  return (
+    <Container>
+      <ImageHolder>
+        <Image src={imageUrl} alt="curratedContentImg" />
```
glenngijsberts: I think this `alt` should be a prop, because I think `curratedContentImg` is not really descriptive for the image itself.

src/components/CurratedContentCard/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  }
+`
+
+interface Props {
+  imageUrl: string
+  title: string
+  text?: string
+}
+
+export const CurratedContentCard = ({ imageUrl, title, text }: Props) => {
```
glenngijsberts: I think `CurratedContentCard` is not a good name for something like this. It can be used with currated content, and maybe for TicketSwap it's only for currated content, but looking broader at this it can be used for any content. Maybe something like `ContentCard` or `ImageCard` is a better name?

Also saw that we use this for the Carousel, so if it's for that, I think it shouldn't live on it's own, but it should be located inside the Carousel folder.

src/components/CurratedContentCard/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  display: flex;
+  inline-size: 302px;
+  block-size: 314px;
+  flex-direction: column;
+  cursor: pointer;
+  border-radius: ${radius.lg};
+  overflow: hidden;
+
+  @media ${device.tablet} {
+    inline-size: 639px;
```
glenngijsberts: Depending on if this is being used in a carousel or not, I don't think it's a good idea to give this a fixed size. If we would use this separately, in for example a full width container, it's not spanning the whole width which looks weird probably:

CleanShot 2023-05-05 at 12 53 38@2x

src/components/CurratedContentCard/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+    flex-direction: row;
+  }
+
+  @media ${device.laptop} {
+    &:hover {
+      img {
+        transition: all ${transition};
+        transform: scale(1.1);
+      }
+    }
```
glenngijsberts: This can use some breaking space :)

src/components/CurratedContentCard/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  @media ${device.tablet} {
+    inline-size: 639px;
+    block-size: 244px;
+    flex-direction: row;
+  }
+
+  @media ${device.laptop} {
+    &:hover {
+      img {
+        transition: all ${transition};
```
glenngijsberts: Do you need to transition all? Maybe better to transition only the `scale`?

src/components/CurratedContentCard/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  }
+`
+const Content = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: flex-start;
+  padding: ${space[16]};
+  max-block-size: 127px;
+  max-inline-size: 302px;
```
glenngijsberts: Instead of using sizes, maybe better to use a grid?

src/components/CurratedContentCard/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  @media ${device.tablet} {
+    font-size: ${fontSize[16]};
+  }
+`
+
+interface Props {
+  imageUrl: string
+  title: string
+  text?: string
```
glenngijsberts: In the designs I didn't saw cards without text. Should we make this required?

src/components/CurratedContentCard/index.stories.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import { CurratedContentCard } from '.'
+
+export default {
+  title: 'Components/Data Display/CurratedContentCard',
+}
```
glenngijsberts: Could you create a variable for this object, and export that variable? That fixes a warning 😄

src/components/Carousel/index.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+      />
+    )
+
+    expect(container.firstChild).toBeInTheDocument()
+    expect(getByText(/Explore the best events/i)).toBeInTheDocument()
+    expect(getByText(/Desert festival/i)).toBeInTheDocument()
+    expect(getByText(/Ocean festival/i)).toBeInTheDocument()
+    expect(getByText(/Jungle festival/i)).toBeInTheDocument()
+  })
+})
```
glenngijsberts: Maybe good to also add tests for controlling the carousel?

src/components/Carousel/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+  @media ${device.tablet} {
+    overflow-x: inherit;
+  }
+`
+
+interface Props {
+  title: string
+  id: string
+  subtitle?: string
+  viewAll?: string
```
glenngijsberts: `viewAll` is a bit vague for what it represents I think. Maybe something like `viewAllLabel`?

src/components/Carousel/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+  onClickAll: () => void
```
glenngijsberts: `onClickViewAll` to keep it consistent I would say.

glenngijsberts: In general I don't think we should have a prop for a `view all` button or something like that, because it means that it always needs to be a button with an onClick, while you might want to pass a link instead. Maybe it should be a prop where you can pass a ReactElement yourself?

src/components/CurratedContentCard/index.stories.tsx
bobbonius: Also, this should be in surfaces category

src/components/Carousel/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+}: Props) => {
+  const [activeIndex, setActiveIndex] = useState(0)
+  const amountOfCardsIsOdd = cards.length % 2 === 1
+
+  const scrollToCard = (accumulator: number) => {
+    const carouselNode = document.getElementById(id)
+    const node = carouselNode?.children[accumulator] || null
+
+    if (node === null) return false
+    scrollIntoView(node, {
```
glenngijsberts: This can use some spacing in between.

src/components/Carousel/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+  ...props
+}: Props) => {
+  const [activeIndex, setActiveIndex] = useState(0)
+  const amountOfCardsIsOdd = cards.length % 2 === 1
+
+  const scrollToCard = (accumulator: number) => {
+    const carouselNode = document.getElementById(id)
+    const node = carouselNode?.children[accumulator] || null
+
+    if (node === null) return false
```
glenngijsberts: Is `if (!node)` not enough? 🤔

glenngijsberts:
CleanShot 2023-05-05 at 12 59 05@2x

On mobile, the card in the carousel is not full width, so you see a little bit of spacing in the container, which doesn't look aligned with the sell all button.

glenngijsberts:
Don't forget to export your components in: `src/components/index.ts` 😄

src/components/CurratedContentCard/index.tsx
bobbonius: Maybe: `FeaturedCard`, `GalleryCard`, `PromotionalCard`

glenngijsberts:
I think this carousel looks similar to the `CardCarousel`. Should we get rid of that?

src/components/Carousel/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+    scrollIntoView(node, {
+      behavior: 'smooth',
+      scrollMode: 'if-needed',
+      inline: 'nearest',
+      boundary: carouselNode,
+    })
+  }
```
glenngijsberts, line 103: Spacing in between.

src/components/CurratedContentCard/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+  }
+`
+const Body = styled(SmallText)`
+  color: ${color.foregroundMuted};
+  overflow: hidden;
+
+  @media ${device.tablet} {
+    font-size: ${fontSize[16]};
+  }
+`
```
glenngijsberts: If this needs a font-size change, why not use `Text`?

src/components/CurratedContentCard/index.tsx
Emacdyz: It will be used for the carousel but also outside of a carousel. `ImageCard` could be better I guess then, thanks!

src/components/CurratedContentCard/index.tsx
Emacdyz: oh yeah, that's a good point!

src/components/CurratedContentCard/index.tsx
Emacdyz: yeah it makes it smoother

src/components/CurratedContentCard/index.tsx
Emacdyz: It is possible to have cards without body text - the new API is not yet started but it was discussed that this won't be required

src/components/Carousel/index.tsx
Emacdyz: Yeah I was hesitant but checked with design and the see all is not optional - it should always be there on any carousel or feed items for that matter. And all See all will lead to a page for consistency. So that's why i thought doing it like this.

src/components/CurratedContentCard/index.tsx
Emacdyz: yeah I asked John to give me aspect ratio so I can implement them

> I think this carousel looks similar to the `CardCarousel`. Should we get rid of that?

Yeah I asked - atm it is used once in sirius in the partner page. So thought it was out of scope for now and apparently this partner page will get a revamp soon as well

src/components/CurratedContentCard/index.tsx
Emacdyz: actually the font-size change should not be there

src/components/CurratedContentCard/index.tsx
Emacdyz: Now using aspect-ratio, i think that is best

src/components/Carousel/index.test.tsx
Emacdyz: Like the previous / next you mean? I tried but struggle a bit to check as all card are present in the document but just hidden. Maybe you have some ideas..

src/components/ImageCard/index.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+      imageAlt="intoTheDunesImage"
+      imageUrl="https://s3-eu-west-1.amazonaws.com/ticketswap-public/public/201610/80afc84b-2091-4359-9fd7-9126d684182c.jpeg"
+      title="Into the dunes Festival"
+      text="Whisper of the sand in the dunes, sun on your face, feet burning... find the oasis"
+    />
+  </Container>
+)
+
+const story = {
+  title: 'Components/Data Display/ImageCard',
```
bobbonius:
```suggestion
  title: 'Components/Surfaces/ImageCard',
```

src/components/Carousel/index.tsx
glenngijsberts: But you also need to think from an outsiders perspective. People might use this without the see all. But even if it's not optional, that doesn't block using a ReactElement for this instead. Using a button for something that's supposed be a link is not optimal.

src/components/Carousel/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+  grid-gap: ${space[16]};
+  scroll-behavior: smooth;
+  scroll-snap-type: x mandatory;
+
+  &::-webkit-scrollbar {
+    display: none;
+  }
+
+  @media ${device.tablet} {
+    overflow-x: inherit;
```
glenngijsberts, line 64: What's the purpose of this? If your goal is to "reset" the scroll, I would use `unset` instead.

src/components/Carousel/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+  }
+`
+
+interface Props {
+  title: string
+  id: string
+  subtitle?: string
+  viewAllLabel?: string
+  onClickViewAll: () => void
+  cards: ReactNode[]
```
glenngijsberts: It's better to use an array of `ReactElement` here. A ReactElement can be only a component, a ReactNode can be almost anything (string, numbers, etc.)

src/components/ImageCard/index.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+}
```
glenngijsberts, line 26: Can use spacing between these two things.

bobbonius:
In general I was wondering. Shouldn't this just be a variant of the `<Card />` component. Because now we have a `<Card image="http://url.com"/>` and a `<ImageCard imageUrl="blabla" />`. Why not build it into the current Card, you have the exact same props available, only thing you need to add is a prop like `featured` or `curated` that results a `bool` and transform the css accordingly? Think its confusing to have another card with that name.

glenngijsberts:
> In general I was wondering. Shouldn't this just be a variant of the `<Card />` component. Because now we have a `<Card image="http://url.com"/>` and a `<ImageCard imageUrl="blabla" />`. Why not build it into the current Card, you have the exact same props available, only thing you need to add is a prop like `featured` or `curated` that results a `bool` and transform the css accordingly? Think its confusing to have another card with that name.

I think it's worth it to look into this as well 👍 I also think that the names provided by Bob are a bit better than the ImageCard I suggested.

src/components/Carousel/index.tsx
Emacdyz: Ok you are right, then I think it should be the same for title/subtitle - maybe other wants to add icons or something else on the left too

src/components/Carousel/index.tsx
Emacdyz: The goal is to disable the horizontal scroll with magic pad on desktop and have only the arrow nav instead. If you don`unset` it is still possible to scroll with the pad

src/components/Carousel/index.tsx
Emacdyz: Great! Thanks!

src/components/Card/index.tsx
```diff
@@ -23,74 +25,120 @@ export enum CardSize {
   large = 'large',
 }
 
-export interface CardPropTypes {
+export interface Props {
```
glenngijsberts: I don't think it's a good idea to change this to `Props`, because we export it, then it's too generic.

src/components/Card/index.tsx
```diff
@@ -23,74 +25,120 @@ export enum CardSize {
...
 const Content = styled.div<StyledCardProps>`
-  position: ${props => (props.hasImage ? 'absolute' : 'relative')};
+  position: ${({ hasImage }) => (hasImage ? 'absolute' : 'relative')};
   inset: 0;
   display: flex;
   flex-direction: column;
   justify-content: flex-end;
+
+  ${({ type }) =>
+    type === 'curatedCard' &&
```
glenngijsberts: I don't think `curatedCard` is a good name, `curated` doesn't' really represents what this is right? Maybe it has more to do with horizontal vs vertical.

src/components/Card/index.tsx
```diff
@@ -309,7 +389,8 @@ const BackgroundImage = styled(Image)`
   }
 `
 
-const Card: React.FC<CardPropTypes> = ({
+const Card = ({
```
glenngijsberts: Nice job on trying this out 👍 

Personally I think it's own component is better, I feel like this component becomes too big / hard to read and less maintainable. Maybe we should create a horizontal image card and a vertical image card?

What do you think?

src/components/Card/index.stories.tsx
```diff
@@ -275,6 +282,19 @@ export const Large = () => (
...
 
+export const CuratedCard = () => (
+  <ContainerCuratedCard>
+    <Card
+      type="curatedCard"
+      image="https://s3-eu-west-1.amazonaws.com/ticketswap-public/public/201610/80afc84b-2091-4359-9fd7-9126d684182c.jpeg"
+      title="Into the dunes Festival"
+      text="Whisper of the sand in the dunes, sun on your face, feet burning... find the oasis"
+    />
+  </ContainerCuratedCard>
```
glenngijsberts: This won't work without container, then it will look like this:

CleanShot 2023-05-09 at 15 21 53@2x

glenngijsberts (commented):
I would say for now go for an `ImageCard` component instead of the Card component approach (See 👇). Other question: Do we really need to have this inside Solar? Maybe we should start in Sirius? Then we also have more freedom in using for example `Curated` as name.

About the component itself, it doesn't work really nice on it's own. If you don't add a container, it will look like this:

CleanShot 2023-05-09 at 15 23 43@2x

I think the ImageCard should get rid of all the "fixed" sizing and work more responsive, with a grid for example. If you need to have it smaller in the Carousel for example, you could re-style it there.

PR #1724, opened 2023-04-26
Make small adjustments to the new cover

glenngijsberts:
# Description

Tweaked the new cover a bit with Mark.

## How to test

- Checkout this branch
- `$ yarn storybook`
- Verify the CoverV2 stories are working as expected
- Verify every cover has also an image on mobile

### DIFF
--- a/src/components/CoverV2/Cover.tsx
+++ b/src/components/CoverV2/Cover.tsx
@@ -43,6 +43,7 @@ export interface CoverImages {
 interface CoverBackgroundImageProps {
   images?: CoverImages
   fullHeight: boolean
+  defaultImage?: string
 }
 
 const CoverBackgroundImage = styled.div<CoverBackgroundImageProps>`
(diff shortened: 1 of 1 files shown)


src/components/CoverV2/Cover.tsx
```diff
@@ -190,9 +197,19 @@ export const Cover = ({
   fullHeight = false,
   ...props
 }: CoverProps) => {
+  const defaultImage =
+    images?.desktop ||
+    images?.tablet ||
+    images?.mobilePortrait ||
+    images?.mobileLandscape
```
viktorrenkema, line 204: I get what's happening here and tried to find other ways to do it, but I'm surprised there don't really seem to be smarter ways to do this 😅

viktorrenkema (approved):
👍

src/components/CoverV2/Cover.tsx
glenngijsberts: yeah it's kinda annoying. I also thought about a lot of different approaches, for example adding an extra prop for a default image or if you only want to use a single image like the old cover, but that is something I want to prevent because I want to make it super clear what prop is responsible for the cover image.

PR #1725, opened 2023-04-26
feat(theme): make adjustments to theme colors

glenngijsberts:
# Description

Mark asked me to change these colors, so see title!

## How to test

- Checkout this branch
- `$ yarn storybook`
- Verify stories are working as expected

## Screenshots
Before
CleanShot 2023-04-26 at 15 09 35@2x

After
CleanShot 2023-04-26 at 15 07 57@2x

### DIFF
--- a/src/global-styles.ts
+++ b/src/global-styles.ts
@@ -73,7 +73,7 @@ export const globalStyles = css`
 
     // Action
     --action: var(--earth400);
-    --actionBackground: var(--earth50);
+    --actionBackground: #00b6f01f;
     --onAction: var(--nova);
     --actionFocus: #00b6f052;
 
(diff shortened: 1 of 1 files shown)


src/global-styles.ts
```diff
@@ -73,7 +73,7 @@ export const globalStyles = css`
 
     // Action
     --action: var(--earth400);
-    --actionBackground: var(--earth50);
+    --actionBackground: #00b6f01f;
     --onAction: var(--nova);
```
viktorrenkema, line 77: Are we not making new variables for these, to be consistent with all other colors?

viktorrenkema (approved):

src/global-styles.ts
glenngijsberts: Hmmm good point, will check with Mark!

PR #1727, opened 2023-05-01
Revert the color of lightForegroundMuted

glenngijsberts:
# Description

The new lightForegroundMuted color was white with some opacity, but that doesn't work nice on light (white, light grey) backgrounds because it's not visible. Now reverting to mitigate the issue. Will create another PR to fix the issue properly.


Marruk:
Where are we using the `lightForegroundMuted` on a light background?

PR #1728, opened 2023-05-01
Fix color issues and create story for theme properties

glenngijsberts:
# Description

In a few recent PR's some colors were changed. This was causing some side-effects (see screenshots). This PR fixes the color of the Card subtitle, introduces a storybook file for the theming (so that we can add stories that are using specific theming properties) and this PR reverts the shadow colors since these are the ones mostly causing side-effects. In an upcoming PR I will re-introduce the new colors, together with Design, using the new storybook file to verify colors are adjusted without any issues.

## Changes

- [x] Make sure to use foreground color instead of light foreground color for Card subtitle 
- [x] Introduce storybook file for theming
- [x] Revert new shadow colors to mitigate issue for now

## How to test

- Checkout this branch
- `$ yarn storybook`
- Verify the Card with Subtitle story for the Card component is working as expected
- Verify the Container with Shadow story for theming is working as expected

## Screenshots

Fixes:

CleanShot 2023-05-01 at 15 30 37@2x

CleanShot 2023-05-01 at 14 59 41@2x


viktorrenkema (approved):
🌤️

PR #1730, opened 2023-05-03
feat(Cover): Small design tweaks to CoverV2

Marruk:
Couple of small design tweaks to the CoverV2, putting a screenshot here for good measure:

Screenshot 2023-05-03 at 16 23 54


src/components/CoverV2/Cover.tsx
```diff
@@ -166,6 +177,8 @@ const Graphic = styled.div`
 `
 
 const MetaInfo = styled.div`
+  max-width: 100%;
```
Marruk, line 180: I added this to let the `text-overflow: ellipsis` work in its children (without it the container just grew as large as it needed to be). Let me know if there is a better way.

glenngijsberts (approved):

PR #1731, opened 2023-05-05
Introduce `StylisedIcon` & `Prompt` components

glenngijsberts:
# Description

In the new cover, but also in this new Prompt component we use an icon on a circle background. Because of this and after aligning with Mark I created a new component called `StylisedIcon` which can be used for the Cover component but also for the early mentioned Prompt component that is being introduced in this PR as well. The Prompt component can serve as empty state, but later on perhaps also as error state. Because both only use a default theming I didn't introduce variants yet, but that shouldn't be hard to setup as soon as we need them.

## How to test

- Checkout this branch
- `yarn storybook`
- Check the stories of `Prompt` and `StylisedIcon`
- Check the stories of `CoverV2` and verify it looks like expected

## To be notified

@Marruk


bobbonius:
Should this be a `StylisedIcon`?

https://github.com/TicketSwap/solar/blob/465c8c846799ab2f803174448824228763de558b/src/components/CoverV2/index.stories.tsx#L86-L88

src/components/StylisedIcon/index.tsx
```diff
@@ -0,0 +1 @@
+export { StylisedIcon } from './StylisedIcon'
```
bobbonius, line 1: What do you think of `DecoratedIcon`?

bobbonius (approved):
Nice work 🔥

src/components/StylisedIcon/index.tsx
glenngijsberts: Sounds better to me tbh, WDYT @Marruk?

Marruk: I like it too!

If it's ok with you I would like to merge it like this because it's called `StylisedIcon` in both Figma and the app design systems. I'll put it on my list to rename it everywhere in one go and keep it the same.

PR #1732, opened 2023-05-05
Revert new `actionBackground` color and re-introduce as `actionBackgroundAlpha12` color

glenngijsberts:
# Description

The new `actionBackground` color was causing some style issues, for example for Select components where elements were visible when options were hovered. This PR reverts the new color and re-introduce it as a `actionBackgroundAlpha12` so we can use it for the secondary variant of the button, where it was introduced for.

After talking with Mark we came up with a naming pattern for transparant colors. Calling it name + `Alpha12`, to make it clear that it contains opacity and in this case 0.12. (see https://gist.github.com/lopspower/03fb1cc0ac9f32ef38f4)

## How to test

- Checkout this branch
- `$ yarn storybook`
- Verify everything works as expected

## Screenshots

CleanShot 2023-05-05 at 12 25 46@2x

CleanShot 2023-05-05 at 12 26 20@2x

### DIFF
--- a/src/components/Button/index.tsx
+++ b/src/components/Button/index.tsx
@@ -165,7 +165,7 @@ const StyledButton = styled.button<StyledButtonProps>`
   ${({ variant }) =>
     variant === ButtonVariant.secondary &&
     css`
-      background-color: ${color.actionBackground};
+      background-color: ${color.actionBackgroundAlpha12};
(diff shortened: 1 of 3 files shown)


viktorrenkema (approved):
Nice! 🌈

PR #1741, opened 2023-05-23
Clean-up semantic colors using color-mix

Marruk:
Recently [color-mix ](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix) got supported by all major browsers.

It has the potential to clean up our semantic colour set by basing variants on a smaller set of base colours. [Some more context here](https://ticketswap.slack.com/archives/C04CDGRKZHR/p1684849529305089).

This PR is a POC taking the `actionBackground` and `actionFocus` as examples.

I added two [utility functions ](https://github.com/TicketSwap/solar/pull/1741/files#diff-5a9ad551a7556d6307d690b38037ae144ae6b6b02626c742a94503c8e36ab58b) to keep the code cleaner as the syntax of `color-mix` is a bit long, but those are optional of course.

### DIFF
--- a/src/components/Button/index.tsx
+++ b/src/components/Button/index.tsx
@@ -6,11 +6,13 @@ import {
   fontSize,
   fontWeight,
   lineHeight,
+  opacity,
   radius,
   space,
 } from '../../theme'
 import { css } from '@emotion/react'
 import { Spinner } from '../Spinner'
+import { transparentize, transparentGradient } from '../../utils/colors'
 
 export enum ButtonVariant {
   primary = 'primary',
@@ -67,11 +69,7 @@ const StyledButton = styled.button<StyledButtonProps>`
   justify-content: center;
   font-family: inherit;
   font-weight: ${fontWeight.semiBold};
-  background-image: linear-gradient(
-    to bottom,
-    rgba(255, 255, 255, 0.24),
-    rgba(255, 255, 255, 0)
-  );
+  background-image: ${transparentGradient('currentColor', 'to bottom', .24, 0)};
   line-height: ${lineHeight.solid};
   align-items: center;
 
@@ -86,7 +84,7 @@ const StyledButton = styled.button<StyledButtonProps>`
     inset: -1px;
     background-color: transparent;
     border-radius: calc(${radius.md} + 1px);
-    box-shadow: 0 0 0 ${space[4]} ${color.actionFocus};
+    outline: ${space[4]} solid ${transparentize(color.action, opacity.statusFocusColor)};
     pointer-events: none;
   }
 
@@ -96,11 +94,7 @@ const StyledButton = styled.button<StyledButtonProps>`
   }
 
   &:hover {
-    background-image: linear-gradient(
-      to bottom,
-      rgba(255, 255, 255, 0.48),
-      rgba(255, 255, 255, 0.24)
-    );
+    background-image: ${transparentGradient('currentColor', 'to bottom', .48, .24)};
   }
 
   ${({ disabled }) =>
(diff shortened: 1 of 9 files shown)


glenngijsberts:
CleanShot 2023-05-23 at 15 16 02@2x

Not sure if this is really nice

src/utils/colors/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  
+  const parsedFromOpacity = 100 - (fromOpacity * 100)
+  const parsedToOpacity = 100 - (toOpacity * 100)
+
+  return `linear-gradient(
+            ${direction},
+            color-mix(in srgb, ${parsedColor}, transparent ${parsedFromOpacity}%),
+            color-mix(in srgb, ${parsedColor}, transparent ${parsedToOpacity}%)
+          )`
+}
```
glenngijsberts, line 29: I would put them in their own files, import them here and then export them from `colors/index.ts`.

src/utils/colors/index.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * Utility function to apply transparency to a color (variable) using `color-mix`
+ * @param {string} color - The color to make transparent. Can be a theme variabale, a hex code or "currentColor"
+ * @param {number} opacity - The desired opacity, a number between 0 and 1
+ */
```
glenngijsberts, line 5: I think this should live inside the commit message instead of a comment.

glenngijsberts:
I'm not really sure if we should introduce utility functions like this instead of a set of colors/values defined in the theme 🤔 Could you maybe explain that in your PR description why you would go for this solution?

src/global-styles.ts
```diff
@@ -73,10 +74,8 @@ export const globalStyles = css`
 
     // Action
     --action: var(--earth400);
-    --actionBackground: var(--earth50);
+    --actionBackground: ${transparentize(color.action, opacity.statusBackgroundColor)};
```
glenngijsberts, line 77: Should this always have an opacity? It introduces this bug again:

CleanShot 2023-05-23 at 15 27 21@2x

Marruk: Yeah, will use a new `activeBackground` here instead so we can distinguish the two use-cases

src/utils/colors/index.ts
Marruk: The plus side of this for me was that the tooltip of the function will explain what is needed, which I thought would be helpful for the `opacity` specifically (there are a number of correct ways to pass that through). Any alternative method for that?

Marruk:
> Not sure if this is really nice

Went with a quick solution, we can do anything.

> I'm not really sure if we should introduce utility functions like this instead of a set of colors/values defined in the theme 🤔 Could you maybe explain that in your PR description why you would go for this solution?

I'm not quite sure what you mean. Are you referring to the removal of `actionFocus`? Or the usage of helper utility functions instead of `color-mix` directly?

src/components/Button/index.tsx
```diff
@@ -86,7 +84,7 @@ const StyledButton = styled.button<StyledButtonProps>`
     inset: -1px;
     background-color: transparent;
     border-radius: calc(${radius.md} + 1px);
-    box-shadow: 0 0 0 ${space[4]} ${color.actionFocus};
+    outline: ${space[4]} solid ${transparentize(color.action, opacity.statusFocusColor)};
```
Marruk, line 87: The reason for this change is because `color-mix` doesn't play nicely with css variables + `box-shadow` (yet)

src/utils/colors/index.ts
glenngijsberts: I do like it tbh, but since it's not a practice in my codebase it seems a bit inconsistent to me 🤷 but yeah not sure what the others think.

glenngijsberts:
> Or the usage of helper utility functions instead of `color-mix` directly?

@Marruk this indeed!

Marruk:
> > Or the usage of helper utility functions instead of `color-mix` directly?
> 
> @Marruk this indeed!

Ah, that was just an idea to keep the code a bit cleaner since the syntax of `color-mix` is quite long. Completely optional though :)

PR #1778, opened 2023-08-07
Add `displayLength` prop to toasts

glenngijsberts:
# Description

This will make it possible to have different timeouts/durations for showing toasts. The default is 3 seconds, which is fine for most toasts but sometimes you have a toast that's a bit longer to read and then it would be nice to have a bit longer duration, without making the toast persistent. 

## How to test

- Checkout this branch
- `yarn storybook`
- Go to the toast stories
- Verify the basic one is showing for 3 seconds
- Verify the long displayLength story is showing for 5 seconds

### DIFF
--- a/src/components/Toast/index.tsx
+++ b/src/components/Toast/index.tsx
@@ -7,6 +7,12 @@ import { css } from '@emotion/react'
 
 const duration = 400
 
+export enum ToastDisplayLength {
+  persistent = 'persistent',
+  short = 'short',
+  long = 'long',
+}
+
 const ItemListContainer = styled.ul`
   position: fixed;
   inset-inline-start: 0;
@@ -24,7 +30,7 @@ const ItemListContainer = styled.ul`
 `
 
 export interface ToastProps {
-  persist?: boolean
+  displayLength?: keyof typeof ToastDisplayLength
   leftAdornment?: ReactNode
   children: ReactNode
 }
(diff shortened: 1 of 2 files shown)


Marruk:
To me it would be nicer to make presets for this use-case, leaving the timing open allows for Toasts to be shown for too little time. The "persist" property could also be changed to be one of those duration presets, with these separate properties you could set yourself a persistent toast that times out after 2 seconds (what happens then?).

(if you agree, we could go with "short" -> current value, "long" -> 5 seconds?, "persistent")

glenngijsberts:
> To me it would be nicer to make presets for this use-case, leaving the timing open allows for Toasts to be shown for too little time. The "persist" property could also be changed to be one of those duration presets, with these separate properties you could set yourself a persistent toast that times out after 2 seconds (what happens then?).
> 
> (if you agree, we could go with "short" -> current value, "long" -> 5 seconds?, "persistent")

@Marruk good idea 👍 will have a look!

src/components/Toast/index.tsx
```diff
@@ -234,9 +240,15 @@ function Item({
...
-        timer = setTimeout(cancel, 3000)
+      if (displayLength === ToastDisplayLength.persistent) return
+
+      if (displayLength === ToastDisplayLength.long) {
+        timer = setTimeout(cancel, 5000)
+
+        return
       }
+
+      timer = setTimeout(cancel, 3000)
```
RobVermeer: Nit

```suggestion
      const delay = displayLength === ToastDisplayLength.long ? 5000 : 3000

      timer = setTimeout(cancel, delay)
```

RobVermeer (approved):
⏲️

src/components/Toast/index.tsx
glenngijsberts: Nice, fixed 👍

PR #1792, opened 2023-08-29
Introduce RadioPanel component

diogoredin:
# Description

This component creates a RadioPanel component based on the Radio button component we already have.

## Changes
- [x] This has been done
- [ ] I still need to to this

## How to test
- Checkout this branch
- `$ yarn dev`

## Screenshots
Screenshot 2023-08-30 at 17 35 51
Screenshot 2023-08-29 at 17 00 26

## To be notified
@RobVermeer


src/components/RadioGroup/Radio.tsx
```diff
@@ -51,13 +51,17 @@ const StyledInput = styled.input`
   }
 `
 
-export type RadioProps = InputHTMLAttributes<HTMLInputElement>
+export interface RadioProps extends InputHTMLAttributes<HTMLInputElement> {
+  $loading?: boolean
```
RobVermeer: Why the dollar sign here? 🤔 

Ahh I think I know why 😄 

You should destruct the props like this
```tsx
({ loading, ...props }: RadioProps, ref: Ref<HTMLInputElement>)
```
so the input just gets the attributes it needs and the render can be clean as well =) (without `props.loading`)

src/components/RadioPanel/RadioPanel.tsx
```diff
@@ -0,0 +1,65 @@
+import React, { ReactNode, Ref, forwardRef } from 'react'
```
RobVermeer, line 1: What do you think of adding this radio panel to the radio folder? Similar to the `RadioWithLabel` or `RadioGroup` (that we export), it's a implemenation of the `Radio` component

src/components/RadioPanel/RadioPanel.tsx
```diff
@@ -0,0 +1,65 @@
...
+  onChange: (event: React.ChangeEvent<HTMLInputElement>) => void
+}
+
+const Wrapper = styled.span`
+  label {
+    display: grid;
+    grid-template-columns: ${space[24]} 1fr;
+    -webkit-align-items: center;
+    -webkit-box-align: center;
+    -ms-flex-align: center;
```
RobVermeer: If I'm not mistaken the browser prefixes will be added automatically, of needed. I also don't see any other component do it.

src/components/RadioPanel/RadioPanel.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { Radio, RadioProps } from '../RadioGroup/Radio'
+import { fontWeight, space } from '../../theme'
+import { Label } from '../Input'
+
+export interface RadioOption
+  extends Omit<RadioProps, 'id' | 'value' | 'label'> {
+  id: string
+  value: string
+  label: ReactNode
+  description?: () => ReactNode
```
RobVermeer: Why did you make this a function? Instead of just `ReactNode` 🤔

src/components/RadioPanel/RadioPanel.tsx
```diff
@@ -0,0 +1,65 @@
...
+}
+
+export interface RadioPanelProps {
+  options: RadioOption[]
+  value: string
+  name: string
+  loading: string
+  onChange: (event: React.ChangeEvent<HTMLInputElement>) => void
```
RobVermeer: You can import `ChangeEvent` named, so we omit `React.`

src/components/RadioPanel/RadioPanel.tsx
```diff
@@ -0,0 +1,65 @@
...
+    -webkit-align-items: center;
+    -webkit-box-align: center;
+    -ms-flex-align: center;
+    align-items: center;
+    gap: ${space[16]};
+    font-weight: ${fontWeight.semiBold};
+  }
+`
+
+export const RadioPanel = forwardRef(
```
RobVermeer: What is the reason for the ref? I think we removed it from all the "v2" components, since we didn't find a good reason for it to use it, but you might have one 😄

src/components/RadioPanel/RadioPanel.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <Panel>
+      {options.map(({ description: Description = () => null, ...option }) => (
+        <PanelContent key={option.id}>
+          <PanelBody>
+            <Wrapper>
+              <Label htmlFor={option.id}>
+                <Radio
+                  name={name}
+                  checked={option.value === value}
+                  ref={ref}
```
RobVermeer: Can a ref be used on multiple elements? 🤯

src/components/RadioPanel/RadioPanel.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }
+`
+
+export const RadioPanel = forwardRef(
+  (
+    { name, value, options = [], loading, onChange }: RadioPanelProps,
+    ref: Ref<HTMLInputElement>
+  ) => (
+    <Panel>
+      {options.map(({ description: Description = () => null, ...option }) => (
```
RobVermeer: If description is just a ReactNode, you can show it by just doing `{description}` and if it's undefined, nothing will be shown =)

src/components/RadioPanel/index.tsx
RobVermeer, line 1: This file doesn't have React, so can be a `.ts` file =)

src/components/RadioPanel/RadioPanel.tsx
RobVermeer, line 1: What do you think of adding this radio panel to the radio folder? Similar to the RadioWithLabel or RadioGroup (that we export), it's a implemenation of the Radio component

src/components/RadioPanel/index.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+  margin-block-end: ${space[16]};
+`
+
+export const Basic = () => {
+  const [value, setValue] = useState('')
+  const [loading, setLoading] = useState('')
+
+  return (
+    <RadioPanel
+      name="Refund Protection"
```
RobVermeer: Let's try to make these stories a bit less TS specific 😝

src/components/RadioGroup/Radio.tsx
diogoredin: it was not to include it in the radio html input, fixed 🙏

src/components/RadioPanel/RadioPanel.tsx
diogoredin: Fixed!

src/components/RadioPanel/RadioPanel.tsx
diogoredin: No need indeed, fixed!

src/components/RadioPanel/RadioPanel.tsx
diogoredin: Changed

src/components/RadioPanel/RadioPanel.tsx
diogoredin: No need, thought we needed for customization, removed 🙏

src/components/RadioPanel/RadioPanel.tsx
diogoredin: True! fixed

src/components/RadioPanel/index.tsx
diogoredin: fixed

src/components/RadioPanel/RadioPanel.tsx
diogoredin: We only actually export the RadioGroup. I made it so you can still import it as a standalone component. Let me know if you had a different idea.

src/components/RadioPanel/index.stories.tsx
diogoredin: Changed 🤪

src/components/RadioPanel/RadioPanel.tsx
diogoredin: removed!

package.json
```diff
@@ -9,7 +9,7 @@
   "types": "./dist/index.d.ts",
   "sideEffects": false,
   "engines": {
-    "node": ">=18"
+    "node": ">=10"
```
RobVermeer: I guess something went wrong with rebasing? 🤓

src/components/RadioGroup/index.stories.tsx
```diff
@@ -31,3 +52,78 @@ export const Basic = () => {
...
   )
 }
+
+export const WithPanel = () => {
+  const [value, setValue] = useState('')
+  const [loading, setLoading] = useState('')
+
+  return (
+    <RadioPanel
+      name="Refund Protection"
```
RobVermeer: This name with the options is gold 😂

RobVermeer (approved):

package.json
diogoredin: Storybook was not working for me with Node 18, so I was temporarily using a lower version (and committed by mistake). Will take a look and open a PR if I have some time.

src/components/RadioGroup/index.stories.tsx
diogoredin: oops 🤪

PR #1824, opened 2023-10-17
fix: build step

diogoredin:
This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. We should use
@babel/plugin-transform-object-rest-spread instead.


RobVermeer (approved):
💠

